Issue #47, opened 2024-06-19
Deploy FastAPI Application


Teneroy:
**Description**

Setup and deploy the Kyma Companion core application using FastAPI, ensuring a structured project setup and foundational components for a robust and scalable application.

**Subtasks**

1. **Define the Project Structure:** Organize the project's directory and file structure for clarity and scalability.
2. **Introduce Package Management System:** Set up Poetry for dependency management and project configuration.
3. **Setup a Basic API Endpoint:** Implement a basic "Hello World" endpoint to verify initial setup.
4. **Setup Basic Logging Infrastructure:** Configure logging to track and debug the application's behavior.
5. **Introduce and Implement Testing Frameworks:** Integrate testing frameworks
6. **Introduce Linters:** Set up code linters to maintain code quality and consistency.
7. **Setup Pydantic:** Implement Pydantic for data validation and settings management.
8. **Build Configuration:** Create and configure a build file (e.g., Dockerfile) to containerize the Kyma Companion application.

**Acceptance Criteria**

- [x] Project structure is clearly defined and organized.
- [x] Poetry is set up and managing project dependencies.
- [x] A basic "Hello World" API endpoint is implemented and functional.
- [x] Logging infrastructure is in place and operational.
- [x] Testing frameworks are integrated, and initial tests are written and passing.
- [x] Linters are configured and ensuring code quality.
- [x] Pydantic is set up and used for data validation
- [x] One can build an image out of this application

PR #51, opened 2024-06-27
Create initial linting, apis, dockerfile, tests

muralov:
**Description**
Creates initial linting configuration with Ruff and implements the inital Dockerfile, tests, and pydantic usage

- Initial REST Apis
- Creates Ruff linting config and enables it
- Initial Dockerfile
- Dummy tests with pytest
- Uses Pydantic
- Documents the executing linting with Ruff

**Related issue(s)**
#47


config/models.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "name": "gpt-4o",
+    "deployment_id": "23590sojsglsfsfsfdsjgls21592jfs"
```
Teneroy, line 4: Let's not hardcode deployment_id in json and publish it into an open-source repo. I think we can load them as env variables, since that would be more secure and flexible approach. In this case we will also have all credentials for llms handled in a similar manner

src/routes/api.py
```diff
@@ -1,12 +1,36 @@
...
 
-@app.get("/chat")
-async def chat():
+def init_chat() -> dict:
+    """ Initialize the chat """
+    logger.info("Initializing chat...")
+    return {"message": "Chat is initialized!"}
+
+
+def chatting() -> dict:
```
Teneroy, line 17:
```suggestion
def process_chat_request() -> dict:
```

src/routes/api.py
```diff
@@ -1,12 +1,36 @@
...
+from src.utils.models import get_model
 
 logger = get_logger(__name__)
 
 app = FastAPI()
 
 
-@app.get("/chat")
-async def chat():
+def init_chat() -> dict:
```
Teneroy, line 11: I am a bit concerned of putting the functions which contain the business logic into an API file. Maybe we can introduce the service layer and move them there? I don't have a strong opinion about it, just think, please, whether it would be a better option

Teneroy (changes requested):

src/routes/api.py
muralov: I think, we should use the async routes to be able to process multiple requests at the same time as  using async functions allows an API to handle concurrent requests more efficiently, especially for I/O-bound operations.

Agree, introducing service layer is good thing. 
It was just dummy routes.

config/models.json
muralov: They will not be hard-coded, we will provide them as JSON in a k8s ConfigMap that is mounted to our application k8s deployment/pod. As we agreed before, I find this better as adding a new model doesn't need any extra env var and deleting doesn't need to delete env vars. I find this approach is more flexible than adding every model name and deployment ID as an env var.

src/routes/api.py
```diff
@@ -1,12 +1,20 @@
...
+
+from src.utils.logging import get_logger
+from src.services.chat import init_chat, process_chat_request
 
 logger = get_logger(__name__)
 
 app = FastAPI()
 
 
+@app.get("/chat/init")
```
mfaizanse, line 11: There is a [official blog](https://fastapi.tiangolo.com/tutorial/bigger-applications/) from FastAPI about defining routes if they are in separate files. Shall we follow that guidelines?

Teneroy (approved):

src/routes/api.py
muralov: I will consider this in the next PR.

PR #52, opened 2024-07-01
Disable testing and release

the1bit:
**Description**

- Pipelines are prepared for new project structure
- e2e pipeline is disabled
- release pipeline is disabled
- main branch modified to `main-1`

**Related issue(s)**

### DIFF
--- a/.github/workflows/backend-relesase.yaml
+++ b/.github/workflows/backend-relesase.yaml
@@ -8,7 +8,7 @@ on:
     branches: ["main"]
     tags: ["*.*.*"]
     paths:
-      - "assistant/backend/**"
+      - "src/**"
 
 # Define the environment variables
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/backend-relesase.yaml
```diff
@@ -5,10 +5,10 @@ name: Backend - release
 
 on:
   push:
-    branches: ["main"]
+    branches: ["main-1"]
```
muralov, line 8: do you really have to do this or is it by mistake?

Teneroy (approved):

.github/workflows/backend-relesase.yaml
the1bit: I did it on purpose. we need to rename it in the future, when we are done with the full restructuring.

Issue #48, opened 2024-06-19
Implement Controllers and Session Management


Teneroy:
**Description**

Develop and integrate necessary controllers for communication with external services and implement session management to maintain user sessions effectively in the Kyma Companion core application. The task depends on #47 . The task can be done after we connect to the redis client. The redis setup depends on this task #25 .

**Subtasks**

1. **Setup Controllers:**
- Implement necessary controllers to enable communication with external services (dummy API, "Hello World" responses with correct data types).
- /chat: Handle conversation with an agent. At the moment just dummy response stream.
- /followups: Generate follow-up questions. Just dummy response.
- /init: Generate initial questions and initialize the conversation with an agent. Just dummy response.
- Ensure controllers can handle multiple requests from different users simultaneously.
2. **Integrate Session Management:**
- Set up the connection to the Redis client for session storage.
- Implement simple session handling with LangGraph [threads](https://langchain-ai.github.io/langgraph/concepts/low_level/#threads)

**Acceptance Criteria**

- [x]  Controllers are implemented and can communicate effectively with necessary external services.
- [x]  /chat controller handles conversations with agents correctly.
- [ ]  /followups controller generates follow-up questions as expected.
- [x]  /init controller generates initial questions and initializes conversations correctly.
- [ ]  Session management is functional and can handle multiple users and sessions.
- [ ]  Connection to Redis client is established and operational.
- [ ]  Middleware for session management is in place and working correctly.

PR #53, opened 2024-07-02
refactor: improve using fastapi and fix running the app locally

muralov:
**Description**
Improve using FastApi and fix running the app locally

- Improve FastApi to run with cli and IDEs
- Add documentation for running and debugging with fastapi

**Related issue(s)**
#48


src/main.py
```diff
@@ -1,13 +1,18 @@
...
+@app.get("/chat/init")
+async def init() -> dict:
+    """ Endpoint to initialize the chat with the Kyma companion """
+    return await init_chat()
+
+
+@app.get("/chat")
+async def chat() -> StreamingResponse:
+    """ Endpoint to chat with the Kyma companion """
+    return StreamingResponse(process_chat_request(), media_type='text/event-stream')
```
mfaizanse, line 18: Why don't we keep the routes separately using the `APIRouter` class? and in main we can include the route with `/api/v1` prefix, so we can easily later version the APIs and also introduce new set of APIs.

mfaizanse (approved):

mfaizanse (approved):

PR #66, opened 2024-07-15
github action linting

grischperl:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/.github/workflows/lint-code.yml
+++ b/.github/workflows/lint-code.yml
@@ -1,0 +1,42 @@
+name: Lint Python code
+
+on:
+  pull_request:
+    branches:
+      - "main"
+      - "release-**"
+    paths-ignore:
+      - "docs/**"
+      - "**.md"
+      - CODEOWNERS
+      - LICENSE
+
+permissions:
+  contents: read
+
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  pull-requests: read
+  # Optional: Allow write access to checks to allow the action to annotate code in the PR.
+  checks: write
+
+jobs:
+  code-linter:
+    name: "Run mypy linting"
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Python # needs to be changed to python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.12'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install mypy
+
+      - name: Run mypy
+        run: mypy .


ruff.toml
```diff
@@ -56,7 +56,7 @@ select = [
     "PLR0915", # keeping functions and methods short and focused
 ]
 
-[per-file-ignores]
+[lint.per-file-ignores]
```
muralov, line 59: why do you need to add `lint` prefix?

pyproject.toml
```diff
@@ -47,6 +51,9 @@ testpaths = [
 
 [tool.poe.tasks]
 lint = "ruff check ."
+typecheck = "mypy ."
```
muralov, line 54: 👍🏼

.github/workflows/lint-code.yml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.12'
```
muralov, line 34: I think it is better if take the python version from our pyproject.toml file. You can change the python version in toml file to a 3.12.2:
```
      - name: Extract Python version from pyproject.toml
        id: python-version
        run: |
          echo "PYTHON_VERSION=$(cat pyproject.toml | grep '^python = ' | sed 's/python = \"//g' | sed 's/\"//g')" >> $GITHUB_ENV
      
      - name: Setup Python
        uses: actions/setup-python@v2
        with:
          python-version: ${{ env.PYTHON_VERSION }}
```

pyproject.toml
```diff
@@ -47,6 +51,9 @@ testpaths = [
...
+format = "black . --check"
+format-fix = "black ."
```
muralov, line 56: I would suggest to have a single task to run the all (mypy, black and linting). It makes it easier to run them locally with a single command, which is a bit convenient and don't forget to run the others

ruff.toml
grischperl: Because of a deprecation warning, I moved to the new prefix:
```
The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `ruff.toml`:
  - 'mccabe' -> 'lint.mccabe'
  - 'pylint' -> 'lint.pylint'
  - 'per-file-ignores' -> 'lint.per-file-ignores'
  ```

pyproject.toml
grischperl: added a slinge command to run all tasks
kept it separated in the github action, but open for discussion

.github/workflows/lint-code.yml
grischperl: done

muralov (approved):

Issue #28, opened 2024-06-18
Automate Creation of Release Artifacts (backend)


Teneroy:
**Description:**
To improve the efficiency and reliability of our development process, we propose the automation of creating release artifacts before the beginning of each development cycle. This will ensure that all necessary components and documentation are ready and available for the development team, reducing delays and increasing productivity. It should adhere to the new building process introduced by neighbors (Google Artifact Registry)

We have a[ backend release process](https://github.com/kyma-project/kyma-companion/blob/main/.github/workflows/backend-relesase.yaml) already implemented, however, this is an old version. We want to use the new standards introduced by neighbors and subsequently build and release our new version of backend and not the POC we developed.

**Acceptance Criteria:**

- [x] A detailed specification document outlining the necessary release artifacts should be created
- [x] Develop a script or set of scripts that can automatically generate the specified release artifacts.
- [x] Integrate the artifact creation script into the CI pipeline to run automatically
- [x] Conduct a trial run to ensure the automation process works as expected and refine as necessary.
- [x] Update internal documentation to include instructions on how the automation process works and how to access the generated artifacts.
- [x] Make sure that the pipeline adheres to currrent SAP standarts(follow guidelines outlined by neighbors as close as possible)

**Recommendation:**
2 people should be working on this task, discuss, and separate responsibilities

the1bit:
Neighbors presented a GitHub based image builder, which is usable for Kyma modules.

Links: 
- https://github.com/kyma-project/test-infra/pull/10983
- https://github.com/kyma-project/test-infra/blob/main/.github/workflows/pull-image-builder-test.yml
- https://github.com/kyma-project/test-infra/blob/main/.github/workflows/pull-build-oidc-token-verifier.yml

Support from Neighbors: @dekiel (https://sap.enterprise.slack.com/team/W8VAMJUQN)

mfaizanse:
**Release pipeline:**

1. Get user inputs:
   - Release version i.e. `x.y.z`
   - Release branch i.e `release-x.y`
2. Validate that version and branch name matches like x.y.z == release-x.y.
3. Check if release tag already exists.
4. Checkout branch
5. Run unit tests.
6. Find the previous release version and generate Changelog.
7. Create draft release and a git tag.
8. Build container image against the git tag and push it to registry.
9. Verify the container image exists `docker pull`
10. (Optional) Run e2e/integration tests against the container image.
11. Publish the release.

**Re-useable build job:**
1. Get user inputs:
[...]

PR #70, opened 2024-07-18
chore: added release pipeline

mfaizanse:
**Description**

- added release pipeline

**Related issue(s)**

- https://github.com/kyma-project/kyma-companion/issues/28

### DIFF
--- a/hack/scripts/create_scan_config.sh
+++ b/hack/scripts/create_scan_config.sh
@@ -6,14 +6,14 @@
 # ./create_scan_config image temp_scan_config.yaml tag
 # - use when bumping the config on the main branch
 
-FILENAME=${1}
-TAG=${2}
-
 # standard bash error handling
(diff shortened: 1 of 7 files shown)


hack/scripts/create_scan_config.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+# This script has the following arguments:
+# - filename of file to be created (mandatory)
+# - release tag (mandatory)
+# ./create_scan_config image temp_scan_config.yaml tag
+# - use when bumping the config on the main branch
+
+FILENAME=${1}
+TAG=${2}
```
marcobebway, line 10: Please let's check if these input exist or not, and fail if they don't.

mfaizanse: updated the script. `set -o nounset` will complain if any arg or var is not defined.

marcobebway (approved):

PR #123, opened 2024-08-12
fix: fix model name for GPT4o

mfaizanse:
**Description**

- fix model name for GPT4o

**Related issue(s)**


config/config.yml
```diff
@@ -1,6 +1,6 @@
 models:
   - name: gpt-4o
-    deployment_id: da6c36722292c60
+    deployment_id: d8cc02e45ca77234
```
muralov, line 3: I think we shouldn't expose this outside?

muralov (approved):

PR #124, opened 2024-08-12
chore: update deployment file for companion

mfaizanse:
**Description**

- update deployment file for companion

**Related issue(s)**

### DIFF
--- a/src/utils/config.py
+++ b/src/utils/config.py
@@ -3,6 +3,8 @@ import os
 import yaml
 from pydantic import BaseModel
 
+from utils.logging import get_logger
+
 
 class Model(BaseModel):
     """Model for the deployment request"""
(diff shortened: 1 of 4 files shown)


src/utils/config.py
```diff
@@ -25,6 +27,7 @@ def get_config() -> Config:
     """
 
     config_file = os.environ.get("MODELS_CONFIG_FILE_PATH", "config/config.yml")
+    get_logger("config").info(f"Loading models config from: {config_file}")
```
muralov, line 30: nitpick: maybe init the logger once outside the class after the imports. We may use it for other functions.

mfaizanse: updated

muralov (approved):

PR #134, opened 2024-08-16
chore: integrate evaluation tests in github action

mfaizanse:
**Description**

- integrate evaluation tests in github action

**Related issue(s)**


.github/workflows/pull-evaluation-tests.yaml
```diff
@@ -0,0 +1,210 @@
...
+# - GARDENER_TEST_CLUSTER_URL
+# - GARDENER_TEST_CLUSTER_AUTH_TOKEN
+# - GARDENER_TEST_CLUSTER_CA_DATA
+
+on:
+  pull_request_target:
+    branches:
+      - main
+    paths:
+      - "src/**"
```
muralov, line 16: I think, the evaluation must run if its logic changes. Should we include evaluation related paths to the GH PR event?

.github/workflows/pull-evaluation-tests.yaml
```diff
@@ -0,0 +1,210 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Setup python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
```
muralov, line 39: nitpick: it'd be better if we use the same python version if possible.

muralov (changes requested):
in general, it looks good. 
I left two comments, one is minor comment.

.github/workflows/pull-evaluation-tests.yaml
mfaizanse: This python version is not for the evaluation tests, but for the python script to wait for build job. The evaluation tests setup uses the same python version defined in the poetry project. https://github.com/kyma-project/kyma-companion/blob/bd64043e2436dff44ed175885fe8a50d1e7aeef0/.github/workflows/pull-evaluation-tests.yaml#L154

.github/workflows/pull-evaluation-tests.yaml
mfaizanse: Its `pull_request_target` so the evaluation tests will run from main branch. So if changes are made to evaluation tests logic we have to test it after the PR is merged.

muralov: Is it not better to run evaluation test scripts to check if scripts have issues?
We run test pipelines if we change a unit test for example.

mfaizanse: updated

.github/workflows/pull-evaluation-tests.yaml
mfaizanse: lets keep it fixed for now.

muralov (approved):

Issue #112, opened 2024-08-07
Create Initial Version of Supervisor Agent for Kyma Companion


Teneroy:
### Description

The goal of this task is to create the initial version of the supervisor agent for Kyma Companion. This includes expanding the chat controller to accept necessary information, setting up a proper technique for the supervisor agent, and creating sample nodes for two agents and one tool for future development.

### Subtasks

**1. Expand Chat Controller:**

- Modify the chat controller to accept session ID and cluster access information from the busload backend.
- Ensure the chat controller can properly handle and store this information for use by the supervisor agent.

**2. Set Up Supervisor Agent Technique:**
Select and set up a suitable technique for the supervisor agent (for example Planning, Routing).

**3. Create Sample Nodes:**

- Develop dummy nodes for two agents that will be used by the supervisor agent. (Kyma agent, Kubernetes agent)
- Develop a dummy node for one tool that will be utilized in future development. (Response conversion tool)

### Acceptance Criteria

- [x] The chat controller is expanded to accept session ID and cluster access information from the busload backend.
- [x] The supervisor agent technique (for example Planning, Routing) is set up.
- [x] Dummy nodes for two agents are created and functional.
- [x]  Planner should consider conversation history. 
   - If possible, supervisor/planner should answer the common questions.

Teneroy:
https://github.com/langchain-ai/langgraph/blob/main/examples/multi_agent/agent_supervisor.ipynb

Teneroy:
Make planner more intelligent by adding information from the state like conversation history and/or messages
Improve Finalizer so it would be able to analyze whether responses from agents make sense, and fail accordingly when they are not
Merge planner and common node for common question. In case there is no need to go with the plan, planner should be able to generate the response to user directly without going to the task executor

muralov:
Improvements in the code:
- [ ] Upgrade generative-ai-hub-sdk package
- [ ] Improve route function according the [discussion](https://sap-btp.slack.com/archives/C07FWF4LFKM/p1724927386684859)
- [ ] Add unit tests for KymaGraph

Teneroy:
Proper streaming format

muralov:
Things that can be improved in the future:
    - sometimes supervisor is failing with output parser error
    - error handling can be improved with retry and fallback logic
    - retry logic is introduced for each node

muralov:
current LangGraph: 
[...]

PR #140, opened 2024-08-20
feat: introduce plan and route with supervisor

muralov:
**Description**
Introduces first implementation of Plan and Route features of Supervisor Agent.

- Implements supervisor agent
- Introduces plan and route feature for supervisor agent
- Introduces common node to answer general questions
- Introduces finalizer node to generate final response
- Introduces generate LangGraph diagram script
- Implements unit tests

**Related issue(s)**
#112


scripts/python/generate_langgraph_diagram.py
```diff
@@ -0,0 +1,36 @@
+import os
+
+from dotenv import load_dotenv
+
+from agents.graph import KymaGraph
+
+load_dotenv()
```
mfaizanse, line 7: Can you write comments on how to run this script and what env variables are needed to run this script?

src/agents/common/agent.py
```diff
@@ -0,0 +1,14 @@
+from typing import Protocol
+
+
+class Agent(Protocol):
```
mfaizanse, line 4: Can you add `Interface` postfix to all interfaces names everywhere? like `AgentInterface` ?

src/agents/common/state.py
```diff
@@ -0,0 +1,65 @@
...
+    result: str | None
+
+    def complete(self) -> None:
+        """Update the result of the task."""
+        self.status = SubTaskStatus.COMPLETED
+
+
+# After upgrading generative-ai-hub-sdk we can message that use pydantic v2
+# Currently, we are using pydantic v1.
+class UserInput(BaseModel):
```
mfaizanse, line 32: This is duplicate of `agents.common.data.Message`. Why we need it?

src/agents/common/state.py
```diff
@@ -0,0 +1,65 @@
...
+
+class AgentState(BaseModel):
+    """Agent state."""
+
+    input: UserInput | None = Field(
+        description="user input with user query and resource(s) contextual information"
+    )
+
+    messages: Annotated[Sequence[BaseMessage], operator.add]
+    next: str | None
```
mfaizanse, line 50: Can you add comments to description `next`?

src/agents/common/state.py
```diff
@@ -0,0 +1,65 @@
...
+    subtasks: list[SubTask] | None = []
```
mfaizanse, line 51: nitpick: `sub_tasks`?

src/agents/common/state.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def all_tasks_completed(self) -> bool:
+        """Check if all the sub-tasks are completed."""
+        return all(task.status == SubTaskStatus.COMPLETED for task in self.subtasks)
+
+
+class Plan(BaseModel):
+    """Plan to follow in future"""
+
+    subtasks: list[SubTask] = Field(
```
mfaizanse, line 63: nitpick: sub_tasks?

src/agents/common/utils.py
```diff
@@ -0,0 +1,88 @@
...
+from langchain_core.messages import AIMessage, BaseMessage, SystemMessage
+from langchain_core.prompts import MessagesPlaceholder
+from langgraph.constants import END
+
+from agents.common.state import AgentState
+from utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+EXIT = "Exit"
```
mfaizanse, line 15: Can we move all global constants to one file `agents/common/constants.py`? Atm they are all dispersed in different files.

src/agents/common/state.py
muralov: yes, there is langchain problem with Pydantic. Generative ai hub sdk needs little older langchain which only work with pydantic v1. This blocked to use pydantic v2. I had to duplicate at the moment. As soon as we upgrade generative ai hub sdk and langchain that supports latest pydantic versions. We can directly use Message class using pydantic v2.
FastAPI with Depends didn't work with pydantic v1. Hence, I was in deadlock and introduced this workaround. 

Let's upgrade generative ai hub sdk in a next step as they introduce huge changes. I haven't trieid it yet.

src/agents/common/agent.py
muralov: I can do that, but it is really discouraged to do so. 
Actually Agent, KymaAgent, SupervisorAgent, KubernetesAgent really make sense if you thing from domain perspective.

supervisor_agent.py
```diff
@@ -0,0 +1,233 @@
+import functools
```
mfaizanse, line 1: Is this file supposed to be merged to main? 
If not, please delete it.
If yes, please move to `/examples` or `/hack` directory.

src/agents/supervisor/agent.py
```diff
@@ -1,128 +1,115 @@
...
-                llm,
-                self.tools,
-                "You are Kyma expert. You assist users with Kyma related questions.",
-            ),
-            name="KymaAgent",
+    """Supervisor agent class."""
+
+    model: Model
+    _name: str = SUPERVISOR
+    parser = PydanticOutputParser(pydantic_object=Plan)  # type: ignore
```
mfaizanse, line 25: Is this parser used anywhere?

src/agents/common/agent.py
mfaizanse: https://dev.to/alvesjessica/interfaces-in-python-not-truly-an-interface-just-a-convention-43d3

Lets atleast for interface, follow the convention of adding I prefix. PEP8 does not talk about it. I see different people are using different convention over the internet.

We do not need to do any prefixes for other types like Enums etc. I would agree with the convention of Agent  -> KymaAgent | K8sAgent if we are doing inheritance. But IMO interfaces are different and they should be clearly marked as abstract.

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def should_continue(state: MessagesState) -> Literal["action", "__end__"]:
+    """Return the next node to execute."""
+    last_message = state["messages"][-1]
+    return "action" if last_message.tool_calls else "__end__"
+
+
+class CustomJSONEncoder(json.JSONEncoder):
+    """Custom JSON encoder."""
```
mfaizanse, line 36: Can you write more comments how does this transfer the obj from which structure to which structure?

mfaizanse: The expected input and output structure.

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+                "error": str(e),
+                "messages": [
+                    AIMessage(
+                        content=f"Error in planning: {e}",
+                        name="Planner",
+                    )
+                ],
+            }
+
+    def common_node(self, state: AgentState) -> dict[str, Any]:
```
mfaizanse, line 201: Can you explain what is the purpose of this node?

src/agents/supervisor/agent.py
```diff
@@ -1,128 +1,115 @@
...
+                            "required": ["description", "assigned_to"],
+                        },
+                    },
+                    "next": {"type": "string", "enum": options},
+                },
+                "required": ["subtasks", "next"],
+            },
+        }
+
+        supervisor_system_prompt = ChatPromptTemplate.from_messages(
```
Teneroy, line 54: Same thing here. Let's move prompts out

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+        self.kyma_agent = KymaAgent(model)
+        self.k8s_agent = KubernetesAgent(model)
+        self.supervisor_agent = SupervisorAgent(
+            model, members=[KYMA_AGENT, K8S_AGENT, COMMON]
+        )
+
+        self.members = [self.kyma_agent.name, self.k8s_agent.name, COMMON]
+        self.graph = self._build_graph()
+
+    def _plan(self, state: AgentState) -> dict[str, Any]:
```
mfaizanse, line 77: Why do we have Supervisor and Planner separately? Isn't it Supervisor is supposed to the planning. IMO, we should merge the Supervisor and planner so we have one entity which is really controller and deciding the flow?

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+        return {
+            "messages": [
+                AIMessage(
+                    content="All my subtasks are already completed.",
+                    name=COMMON,
+                )
+            ]
+        }
+
+    def generate_final_response(self, state: AgentState) -> dict[str, Any]:
```
mfaizanse, line 256: Shall we introduce final_response as a separate agent?

src/services/conversation.py
```diff
@@ -0,0 +1,74 @@
...
+    This class is a singleton and should be used to handle the conversation.
+    """
+
+    kyma_graph: Graph
+
+    def __init__(self):
+        model_factory = ModelFactory()
+        model = model_factory.create_model(LLM.GPT4O_MODEL)
+        memory = RedisSaver(
+            async_connection=initialize_async_pool(url=f"{os.getenv('REDIS_URL')}/0")
```
Teneroy, line 58: Let's move Redis to a constant

src/utils/langfuse.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+from langfuse.callback import CallbackHandler
+
+handler = CallbackHandler(
+    secret_key=os.getenv("LANGFUSE_SECRET_KEY"),
```
Teneroy, line 6: Let's move to some constant variables

src/agents/common/utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+DEFAULT_NUMBER_OF_MESSAGES = 10
+
+
+def filter_messages(
+    messages: Sequence[BaseMessage],
+    last_messages_number: int = DEFAULT_NUMBER_OF_MESSAGES,
+) -> Sequence[BaseMessage]:
+    """This is very simple helper function which only ever uses the last four messages"""
```
Teneroy, line 74: The comment is misleading

Teneroy (changes requested):

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+        """Generate the final response."""
+
+        prompt = ChatPromptTemplate.from_messages(
+            [
+                MessagesPlaceholder(variable_name="messages"),
```
Teneroy, line 261: Supervisor agent should have an ability to fail properly in case other agents can't provide relevant information. Therefore it should probably analyze provided information and make a decision whether this information is sufficient and it can generate a comprehensive response or it should fail. I suspect, additional prompting should be done in the finalizer. Here is an example where an agent fails in doing that:
[request_response1.txt](https://github.com/user-attachments/files/16798505/request_response1.txt)

Teneroy (changes requested):

src/agents/supervisor/agent.py
```diff
@@ -1,128 +1,115 @@
...
-) -> dict[str, Any]:
-    """It filters the messages and invokes the agent."""
-    state["messages"] = filter_messages(state["messages"])
-
-    result = agent.invoke(state)
-    return {"messages": [HumanMessage(content=result["output"], name=name)]}
+logger = get_logger(__name__)
 
 
 class SupervisorAgent:
```
Teneroy, line 20: Consider renaming to task executor. Then we will merge planner, task executor, replan, and finalizer into the subgraph called SupervisorAgent

Teneroy (changes requested):

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+            ]
+        ).partial(
+            members=", ".join(self.members),
+            output_format=self.parser.get_format_instructions(),
+        )
+        planner = planner_prompt | self.model.llm
+        try:
+            plan = planner.invoke(
+                {
+                    "messages": state.messages[-1:],
```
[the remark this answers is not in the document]
muralov: We need think whether we should consider the conversation or not for the planning. I am currently not sure.
I will document the points in the issue and improve them further in the next sprints.
Let's keep it simple for the start?

src/agents/graph.py
muralov: if you ask my opinion, I wouldn't do it as the complex work is already done by the specific agents. 
I'll document the point in the issue so that we think together. Is it okay?

src/agents/supervisor/agent.py
muralov: yes, below, you can search for it with `state.subtasks = self.parser.parse(plan.content).subtasks`

src/agents/graph.py
muralov: As we discussed in the architecture meeting, if you have them in a single node it tries to re-plan again if they a single node. To avoid for the beginning I've put it outside. 
In the next iteration we optimize our nodes. 
I'll document this in the issue.

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+from utils.langfuse import handler
+from utils.logging import get_logger
+from utils.models import Model
+
+logger = get_logger(__name__)
+
+PLANNER = "Planner"
+
+
+def should_continue(state: MessagesState) -> Literal["action", "__end__"]:
```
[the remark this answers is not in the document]
muralov: I'll delete it

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+    """Return the next node to execute."""
+    last_message = state["messages"][-1]
+    return "action" if last_message.tool_calls else "__end__"
+
+
+class CustomJSONEncoder(json.JSONEncoder):
+    """Custom JSON encoder."""
+
+    def default(self, obj):  # noqa D102
```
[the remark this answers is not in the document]
muralov: yes. There are mainly two reasons I haven't done it yet:
- first, didn't have enough time
- second, we need to add status code (error or success) for the chunks as anytime error may occur during streaming. The status code in the begining is almost always 200 due asynchronous behavior of FastAPI routes.  

For that reason, I'd suggest to have follow-up ticket for this? We can think how to add the code and implement it in the follow-up ticket.
I'll also document this in the issue.

src/agents/graph.py
```diff
@@ -0,0 +1,342 @@
...
+                        content=f"Error in planning: {e}",
+                        name="Planner",
+                    )
+                ],
+            }
+
+    def common_node(self, state: AgentState) -> dict[str, Any]:
+        """Breaks down the given user query into sub-tasks."""
+
+        prompt = ChatPromptTemplate.from_messages(
```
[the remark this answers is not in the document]
muralov: We need think whether we should consider the conversation or not for the planning. I am currently not sure.
I will document the points in the issue and improve them further in the next sprints.
Let's keep it simple for the start?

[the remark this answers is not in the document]
muralov: if common questions, don't plan and directly send to Finalizer!

src/agents/supervisor/agent.py
muralov: I would leave this as Supervisor as it eventually will have more functionality. Planning is just temporarily separated out for the start. It will be part of SupervisorAgent too soon.

src/agents/graph.py
muralov: it answer the general questions. We'll move this to planner if possible in the next PR.

src/agents/common/utils.py
```diff
@@ -0,0 +1,93 @@
...
+        "messages": [
+            AIMessage(
+                content="All my subtasks are already completed.",
+                name=name,
+            )
+        ]
+    }
+
+
+DEFAULT_NUMBER_OF_MESSAGES = 10
```
mfaizanse, line 65: lets move this to the common file?

src/services/conversation.py
```diff
@@ -0,0 +1,74 @@
...
+
+from agents.common.data import Message
+from agents.graph import IGraph, KymaGraph
+from agents.memory.redis_checkpointer import RedisSaver, initialize_async_pool
+from utils.logging import get_logger
+from utils.models import LLM, ModelFactory
+
+logger = get_logger(__name__)
+
+REDIS_URL = f"{os.getenv('REDIS_URL')}/0"
```
mfaizanse, line 13: Ideally, this should be part of a separate module `Config`.

src/services/conversation.py
```diff
@@ -0,0 +1,74 @@
...
+        self, conversation_id: int, message: Message
+    ) -> AsyncGenerator[bytes, None]:
+        """Handle a request for a conversation"""
+        ...
+
+
+class SingletonMeta(type):
+    """Singleton metaclass."""
+
+    _instances: dict[type, object] = {}
```
mfaizanse, line 35: Why it needs to be a dict? Can it be a single `_instance: object`?

src/agents/common/state.py
muralov: let's leave it as it is as the word `subtask` really exists :)

src/services/conversation.py
muralov: yes, I already have this in mind for the next PR. I will create a TODO for this too.

mfaizanse (approved):

src/services/conversation.py
muralov: this meta class have class field _instances which can be used by the other classes too. It means it stores a single instances for others.

Teneroy (approved):

PR #154, opened 2024-09-11
feat: improve planner and router

muralov:
**Description**
Implements improvements for Supervisor Agent regarding planning, routing and more

- Response with planner for general questions
- Simplify routing 
- ...

**Related issue(s)**
#112


src/agents/supervisor/agent.py
```diff
@@ -24,48 +26,35 @@ class SupervisorAgent:
...
-            )
-            | JsonOutputFunctionsParser()
+        ).partial(
+            options=str(self.options),
+            members=", ".join(self.options),
+            finalizer=FINALIZER,
+            output_format=self.parser.get_format_instructions(),
         )
 
+        return supervisor_system_prompt | self.model.llm  # type: ignore
```
grischperl, line 56: Why can this function either return a ChatPromptTemplate or model.llm?

src/agents/common/utils.py
```diff
@@ -80,14 +74,38 @@ def filter_messages(
     return messages[-last_messages_number:]
 
 
-def should_exit(state: AgentState) -> Literal["exit", "continue"]:
-    """Return the next node based on the state."""
-    return "exit" if state.error else "continue"
+def next_step(state: AgentState) -> Literal[EXIT, FINALIZER, CONTINUE]:  # type: ignore
+    """Return EXIT if there is an error, FINALIZER if the next node is FINALIZER, else CONTINUE."""
```
grischperl, line 78: EXIT is also returned if the next state is EXIT

grischperl:
Could you generate the picture of the graph and add it here? It would help me with reviewing the changes.

src/agents/common/utils.py
```diff
@@ -80,14 +74,38 @@ def filter_messages(
...
+        logger.debug("Ending the workflow.")
+        return EXIT
+    if state.error:
+        logger.error(f"Exiting the workflow due to the error: {state.error}")
+        return EXIT
+    return FINALIZER if state.next == FINALIZER else CONTINUE
 
 
 def exit_node(state: AgentState) -> dict[str, Any]:
     """Used in case of an error."""
```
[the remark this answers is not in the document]
muralov: This is general exit node:
image

src/agents/common/utils.py
```diff
@@ -80,14 +74,38 @@ def filter_messages(
...
+
+
+def create_node_output(
+    message: BaseMessage | None = None,
+    next: str | None = None,
+    subtasks: list[SubTask] | None = None,
+    final_response: str | None = None,
+    error: str | None = None,
+) -> dict[str, Any]:
+    """Create a LangGraph node output."""
```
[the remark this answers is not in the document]
muralov: This is the central function with which a LangGraph node output should be created. It is a helper function not to miss any dictionary field. In the upcoming PRs, we should try to use this everywhere if possible.
I'll add more documentation to it.

src/agents/supervisor/agent.py
muralov: The sign `|` is pipe sign of the langchain where it used to create a langchain chain. It is not a OR sign :)

muralov:
> Could you generate the picture of the graph and add it here?

image

src/agents/common/utils.py
```diff
@@ -62,12 +69,9 @@ def agent_node(state: AgentState, agent: AgentExecutor, name: str) -> dict[str,
     }
 
 
-DEFAULT_NUMBER_OF_MESSAGES = 10
-
-
 def filter_messages(
     messages: Sequence[BaseMessage],
-    last_messages_number: int = DEFAULT_NUMBER_OF_MESSAGES,
+    last_messages_number: int = RECENT_MESSAGES_LIMIT,
```
grischperl, line 74:
```suggestion
recent_messages_limit
```

src/agents/common/utils.py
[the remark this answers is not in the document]
grischperl: Then, the description does not match. The description suggests it's only used in an error case.

src/agents/common/utils.py
```diff
@@ -80,14 +84,47 @@ def filter_messages(
...
+        logger.debug("Ending the workflow.")
+        return EXIT
+    if state.error:
+        logger.error(f"Exiting the workflow due to the error: {state.error}")
+        return EXIT
+    return FINALIZER if state.next == FINALIZER else CONTINUE
 
 
 def exit_node(state: AgentState) -> dict[str, Any]:
     """Used in case of an error."""
```
grischperl, line 99:
```suggestion
"""Node to exit the workflow."""
```

src/agents/graph.py
```diff
@@ -81,40 +109,47 @@ def _plan(self, state: AgentState) -> dict[str, Any]:
...
-                        )
-                    ]
+
+            try:
+                plan = self.plan_parser.parse(plan_response.content)
+                if plan.response:
+                    return create_node_output(
+                        message=AIMessage(content=plan.response, name=PLANNER),
+                        final_response=plan.response,
+                        next=EXIT,
```
grischperl, line 128: This is the case, where the planner does not come up with a plan but instead has a direct response. 

Is the `next=EXIT` correct? Shouldn't this direct response go to the finalizer first?

src/agents/graph.py
```diff
@@ -81,40 +109,47 @@ def _plan(self, state: AgentState) -> dict[str, Any]:
...
+                    )
+            except OutputParserException as ope:
+                logger.debug(f"Problem in parsing the plan: {ope}")
```
grischperl, line 131: nit: `Problem in parsing the planner response`

src/agents/graph.py
muralov: at the moment, we don't need another LLM call as planner already is able to answer the question, which makes it quite fast.
Forwarding it to finalizer creates another LLM call and sometimes creates fuzzy answer.

grischperl: - The way I understood it, all responses should go through the `Finalizer`. Is this not correct?
 - The `Planner` either returns a plan or a direct response. In case of a plan, next comes an agent. In case of a direct response, next comes `Exit`. Why does the generated graph show that the `Planner` can also go to the `Finalizer`?

muralov: - If planner already can answer (e.g. general query) or error happens, it is immediately ending with exit node w/o finalizer.
- in case of plan, the plan is forwarded to supervisor where the agents called and eventually finalizer finalizes the messages of the agents. Finalizer is special for Kyma and K8s.

grischperl (approved):

PR #163, opened 2024-09-12
feat: improve the code and chunk response

muralov:
**Description**
Improve the code and change the chunk response as designed

  - Make chunk response as designed
  - Move all env vars to a single module
  - Document generate_langgraph_diagram python script
  - Add and adapt unit tests

**Related issue(s)**
#112


src/agents/graph.py
```diff
@@ -129,6 +133,7 @@ def _plan(self, state: AgentState) -> dict[str, Any]:
                     )
             except OutputParserException as ope:
                 logger.debug(f"Problem in parsing the planner response: {ope}")
+                # sometimes the plan_response is without 'response' attribute
```
friedrichwilken, line 136: It is an interesting fact that you describe here but can you please write a comment that correlates to what is happening in the code like `In case this or that happens we will do this and that.`> In this case maybe something like 'If `plan_response` is missing we do ...'

friedrichwilken (changes requested):

src/agents/graph.py
```diff
@@ -129,6 +133,9 @@ def _plan(self, state: AgentState) -> dict[str, Any]:
                     )
             except OutputParserException as ope:
                 logger.debug(f"Problem in parsing the planner response: {ope}")
+                # If 'response' field of the content of plan_response is missing due to LLM inconsistency,
+                # th response is read the from the plan_response content.
```
friedrichwilken, line 137:
```suggestion
                # the response is read the from the plan_response content.
```

src/agents/graph.py
```diff
@@ -129,6 +133,9 @@ def _plan(self, state: AgentState) -> dict[str, Any]:
...
+                "If plan_response is missing we do ..."
```
friedrichwilken, line 138: What is the purpose of this string?

muralov: it was by mistake, will delete it.

friedrichwilken (approved):

PR #174, opened 2024-09-20
feat: apply token limit to initial question generation

friedrichwilken:
**Description**

- To prevent that the prompt that creates the inital questions exceeds the token limit of the used model, this PR will add logic that will trim the used token number down accordingly. 

**Related issue(s)**

### DIFF
--- a/src/initial_questions/inital_questions.py
+++ b/src/initial_questions/inital_questions.py
@@ -1,6 +1,7 @@
 import typing
 from typing import Protocol
 
+import tiktoken
 import yaml
 from langchain_core.prompts import PromptTemplate
 
@@ -28,23 +29,43 @@ class IInitialQuestionsHandler(Protocol):
         """Fetch the relevant data from Kubernetes cluster based on specified K8s resource in message."""
         ...
 
+    def apply_token_limit(self, text: str, token_limit: int) -> str:
+        """Reduces the amount of tokens of a string by truncating exeeding tokens.
+        Takes the template into account."""
+        ...
+
 
 class InitialQuestionsHandler:
     """Handler that generates initial questions."""
 
     chain: typing.Any
+    _model: IModel
+    _template: str
 
-    def __init__(
-        self,
-        model: IModel,
-    ) -> None:
-        prompt_template: str = INITIAL_QUESTIONS_PROMPT
+    def __init__(self, model: IModel, template: str | None = None) -> None:
+        self._model = model
+        self._template = template or INITIAL_QUESTIONS_PROMPT
+
         prompt = PromptTemplate(
-            template=prompt_template,
+            template=self._template,
             input_variables=["context"],
         )
         output_parser = QuestionOutputParser()
         self.chain = prompt | model.llm | output_parser
+
+    def apply_token_limit(self, text: str, token_limit: int) -> str:
+        """Reduces the amount of tokens of a string by truncating exeeding tokens.
+        Takes the template into account."""
+        tokenizer = tiktoken.encoding_for_model(self._model.name)
+
+        tokens_template = tokenizer.encode(self._template)
+        tokens_text = tokenizer.encode(text)
+        token_limit -= len(tokens_template)
+
+        if len(tokens_text) > token_limit:
+            tokens_text = tokens_text[:token_limit]
+
+        return tokenizer.decode(tokens_text)
 
     def generate_questions(self, context: str) -> list[str]:
         """Generates initial questions given a context with cluster data."""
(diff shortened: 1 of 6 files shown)


src/initial_questions/inital_questions.py
```diff
@@ -28,24 +29,44 @@ def fetch_relevant_data_from_k8s_cluster(
...
+    def apply_token_limit(self, text: str, token_limit: int) -> str:
+        """Reduces the amount of tokens of a string by truncating exeeding tokens.
+        Takes the template into account."""
+        tokenizer = tiktoken.encoding_for_model(self._model.name)
+
+        tokens_template = tokenizer.encode(self._template)
+        tokens_text = tokenizer.encode(text)
+        token_limit -= len(tokens_template)
+
+        if len(tokens_text) > token_limit:
```
muralov, line 65: nitpick: can we have something like this instead of subtracting template token limit (w/o subtraction):
```
if len(tokens_text) + len(tokens_template) > token_limit:
```
improves the readability little bit.

src/services/conversation.py
```diff
@@ -19,6 +19,8 @@
 
 logger = get_logger(__name__)
 
+TOKEN_LIMIT = 14_000
```
muralov, line 22: is `14_000` what model approximately can accept at most? Why did you decide for this number?

src/initial_questions/inital_questions.py
```diff
@@ -28,24 +29,44 @@ def fetch_relevant_data_from_k8s_cluster(
...
+            template=self._template,
             input_variables=["context"],
         )
         output_parser = QuestionOutputParser()
         self.chain = prompt | model.llm | output_parser
 
+    def apply_token_limit(self, text: str, token_limit: int) -> str:
+        """Reduces the amount of tokens of a string by truncating exeeding tokens.
+        Takes the template into account."""
+        tokenizer = tiktoken.encoding_for_model(self._model.name)
```
muralov, line 59: can you cache the `tokenizer` instance as an object field as it is initialised for every initial questions call?

src/initial_questions/inital_questions.py
```diff
@@ -28,24 +29,44 @@ def fetch_relevant_data_from_k8s_cluster(
...
+
+        tokens_template = tokenizer.encode(self._template)
+        tokens_text = tokenizer.encode(text)
+        token_limit -= len(tokens_template)
```
muralov, line 63: sometimes tokens_template token number can be higher than tokens_text. In this case `tokens_limit` will be negative. You need to handle that case too.

muralov (changes requested):

src/initial_questions/inital_questions.py
```diff
@@ -28,24 +29,44 @@ def fetch_relevant_data_from_k8s_cluster(
...
-        prompt_template: str = INITIAL_QUESTIONS_PROMPT
         prompt = PromptTemplate(
-            template=prompt_template,
+            template=self._template,
             input_variables=["context"],
         )
         output_parser = QuestionOutputParser()
         self.chain = prompt | model.llm | output_parser
 
+    def apply_token_limit(self, text: str, token_limit: int) -> str:
```
muralov, line 56: maybe later, we somehow enable LangFuse tracing for such methods too. You don't to do it at the moment, if is not appearing in the tracing. Let's do it later with a follow-up ticket.

src/initial_questions/inital_questions.py
friedrichwilken: "Sometimes tokens_template token number can be higher than tokens_text. In this case tokens_limit will be negative. You need to handle that case too."

No.

Say token_template has a len of 10, token_text has a len of 5, so as you say template>text. Let's say TokenLimit is 12.  `token_limit -= len(tokens_template)` would still not be negative. 

It would only be negative in the unlikely event of the template being longer than the limit. If the template is longer than the token limit we will not be able to send a request to the model anyhow because it does not make any sense to shorten the template.

friedrichwilken: I have chosen an in-between way; I will now handle the case where template is longer than token limit.

src/initial_questions/inital_questions.py
friedrichwilken: agree

src/services/conversation.py
friedrichwilken: That number indeed is wrong, I will fix it.
https://platform.openai.com/docs/models/gpt-4o

src/initial_questions/inital_questions.py
friedrichwilken: I have chosen a similar way, because I will have to do multiple len operations.

src/initial_questions/inital_questions.py
friedrichwilken: Good idea.

muralov (approved):

PR #180, opened 2024-09-24
chore: evaluation test only runs if 'evaluation requested' label is present

friedrichwilken:
**Description**

- change the eval test to only run when the `evaluation requested` label is present.

**Related issue(s)**

### DIFF
--- a/.github/workflows/pull-evaluation-tests.yaml
+++ b/.github/workflows/pull-evaluation-tests.yaml
@@ -17,6 +17,7 @@ on:
       - "**.md"
       - CODEOWNERS
       - LICENSE
+    types: [labeled]
 
 # global env variables.
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-evaluation-tests.yaml
```diff
@@ -9,14 +9,11 @@ run-name: "Evaluation tests"
...
-    paths-ignore:
-      - "docs/**"
-      - "**.md"
-      - CODEOWNERS
-      - LICENSE
+  workflow_dispatch:
+    inputs:
+      branch:
+        required: true
+        description: Branch name to trigger the workflow
```
mfaizanse, line 16: How can we manually trigger it on PRs?

friedrichwilken: Via a label check; if that works.

mfaizanse (approved):

Issue #201, opened 2024-10-02
Implement integration tests for k8s client


mfaizanse:
Currently, the [`K8sClient` module](https://github.com/kyma-project/kyma-companion/blob/main/src/services/k8s.py) do not have any tests because unit test would not be enough for this module and it is also a critical module for kyma-companion. We should implement integration tests to ensure the module works as expected.

**Task(s):**
1. Implement integration tests for [`K8sClient` module](https://github.com/kyma-project/kyma-companion/blob/main/src/services/k8s.py) either with `envtests` or `k3d`.
2. Add a Github Action job to run these integration tests on PRs and main + release branches.

PR #217, opened 2024-10-16
chore: added integration tests for K8sClient

mfaizanse:
**Description**

- added integration tests for K8sClient.
- updated integration tests github action.

**Related issue(s)**

- https://github.com/kyma-project/kyma-companion/issues/201


tests/integration/services/test_k8s.py
```diff
@@ -0,0 +1,401 @@
...
+            for item in result["events"]:
+                assert isinstance(item, dict)
+
+    @pytest.mark.parametrize(
+        "given_namespace",
+        [
+            # Test case: should be able to get list of not running pods.
+            "nginx-wrong-image",
+            # Test case: should be able to get list of not running pods in all namespaces.
+            "",
```
muralov, line 254: should we verify in the test logic that it is really retrieving all the pods in all namespace? With pods list size, for example.

muralov (commented):
In general it looks good. 

I understood you are just using the resources already exist in the gardener cluster. 
Should we improve it to create its own test resources and clean up the resource after tests are completed in a follow-up PR?
Moreover, it'd be nice if we can run it with envtest thing so that we can run it locally.

tests/integration/services/test_k8s.py
mfaizanse: Well, I started with that approach with a k3d cluster but realized that its duplicate work which we have already done in evaluation tests. We only need to read data, if we would have needed to update resources then having a separate isolated k8s env would be helpful. We can for now keep this and do improvements in a separate issue after discussing with the team.

tests/integration/services/test_k8s.py
```diff
@@ -0,0 +1,401 @@
...
+from http import HTTPStatus
+
+import pytest
+from dotenv import load_dotenv
+from requests import Response
+
+from services.k8s import K8sClient
+
+# load env file if exists.
+load_dotenv(os.getenv("ENV_FILE", "../../../.env.evaluation"))
```
muralov, line 11: can you load the `.env.test` instead of `env.evaluation`. 
You don't even load it is loaded by the settings.py script.

tests/integration/services/test_k8s.py
muralov: I see, we have to make sure the test data is not changed that may break the tests.

tests/integration/services/test_k8s.py
mfaizanse: I have updated it.

tests/integration/services/test_k8s.py
mfaizanse: I will add documentation in a separate PR.

.gitignore
```diff
@@ -233,7 +233,7 @@ scripts/k8s/companion-configmap.yaml
 commands.md
 .env.dev
 .env.fzn.dev
-tests/blackbox/.env.evaluation
+**/.env.evaluation
```
muralov, line 236: nitpick: would you like to delete/change this too?

muralov (approved):

Issue #215, opened 2024-10-15
Implement Follow-Up Questions Endpoint


Teneroy:
## Description

The goal of this task is to implement the **follow-up questions** endpoint. This endpoint will retrieve conversation history from **Redis Checkpointers** and generate from **1 to 3 follow-up questions** based on the context of the conversation history. The endpoint should return the generated questions in the format agreed upon in the API description document.

## Tasks

1. **Implement Follow-Up Questions Endpoint:**
   - Develop the API endpoint that handles requests for follow-up questions.
   - Ensure the endpoint is compliant with the structure and behavior described in the API documentation.

2. **Retrieve Conversation History from Redis Checkpointers:**
   - Implement logic to retrieve conversation history from **Redis Checkpointers**.
   - Ensure the retrieved history includes necessary information (e.g., user inputs, companion responses) to generate follow-up questions.

3. **Generate Follow-Up Questions:**
   - Develop a function to generate **1 to 3 follow-up questions** based on the conversation history.
   - Use conversation context to ensure the questions are relevant and insightful for the next steps in the interaction.
   - Ensure the questions align with the overall flow of the conversation and make sense based on previous user inputs and companion outputs.

4. **Return Follow-Up Questions in the Correct Format:**
   - Ensure the API response returns the follow-up questions in the correct format, as agreed in the API description document.

## Acceptance Criteria

- [x] Implement the **follow-up questions** endpoint as specified in the API documentation.
- [x] Retrieve conversation history from **Redis Checkpointers**.
- [x] Generate between **1 and 3 follow-up questions** based on the conversation history.
- [x] Ensure the follow-up questions are contextually relevant to the conversation so far.

Teneroy:
Consider using 4o-mini

PR #252, opened 2024-11-05
feat: implement initial version of follow-up questions endpoint

mfaizanse:
**Description**

- implement initial version of follow-up questions endpoint
- Example outputs:
  ```
  How can I extend my applications using Kyma?
  What are the best practices for managing services in Kyma?
  What issues might arise when integrating external services with Kyma?
  What is the role of microservices in Kyma?
  What is the concept of serverless functions in Kyma?
  ```
  
  ```
  What steps can I take to resolve the CrashLoopBackOff?
  What logs should I check for the nginx pod?
  What is the role of the nginx container in Kubernetes?
  How can I improve my pod's stability?
  How can I follow best practices for my pod configuration?"
  ```

**Related issue(s)**

- https://github.com/kyma-project/kyma-companion/issues/215


src/followup_questions/followup_questions.py
```diff
@@ -0,0 +1,92 @@
...
+from utils.models import IModel
+
+logger = get_logger(__name__)
+
+
+class IFollowUpQuestionsHandler(Protocol):
+    """Protocol for IFollowUpQuestionsHandler."""
+
+    def generate_questions(self, messages: list[BaseMessage]) -> list[str]:
+        """Generates initial questions given a context with cluster data."""
```
grischperl, line 27: Is this used for initial q's as well?

src/followup_questions/followup_questions.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def generate_questions(self, messages: list[BaseMessage]) -> list[str]:
+        """Generates follow-up questions given the conversation history."""
+        if len(messages) == 0:
+            return []
+
+        # filter down the conversation history to limit the token count.
+        history = self._get_filtered_history(messages)
+        # invoke the chain to generate follow-up questions.
+        return self._chain.invoke({"history": history})  # type: ignore
```
grischperl, line 65: Is there nothing that can be done to fix this instead of ignoring the type checking error?

tests/integration/agents/test_followup_questions.py
```diff
@@ -0,0 +1,138 @@
...
+shared_memory = {}
+
+
+@pytest.fixture
+def followup_correctness_metric(evaluator_model):
+    return GEval(
+        name="Correctness",
+        evaluation_steps=[
+            "Check whether each line is a valid question",
+            "Check if each questions if related to either Kubernetes or SAP Kyma runtime",
```
grischperl, line 27: Typo

```suggestion
            "Check if each questions is related to either Kubernetes or SAP Kyma runtime",
```

grischperl:
What do you think about defining this PR as a `feat` instead of a `chore`?
I think follow up questions could be considered as a new feature.

src/followup_questions/followup_questions.py
mfaizanse: no, only for followup questions. Initila questions have a separate method with similar name.

src/followup_questions/followup_questions.py
mfaizanse: AFAIK, it cannot be fixed atm.

src/followup_questions/followup_questions.py
grischperl: It says "initial questions" right now

mfaizanse: Good catch! Updated!

grischperl (approved):

PR #271, opened 2024-11-22
Add message filter to only pass human messages, system messages and finalizer messages to planner

friedrichwilken:
**Description**

- add message filter utils
- use message filter utils to filter only pass the 10 most recent messages that are human messages (user input), system messages (brief context overviews) and finalizer messages (companion output)

**Related issue(s)**


tests/unit/utils/test_filter_messages.py
```diff
@@ -0,0 +1,163 @@
...
+            "should return empty list when no messages pass the checks",
+            [
+                HumanMessage(content="Human Message"),
+                AIMessage(content="AI Message"),
+                SystemMessage(content="Sysytem Message"),
+            ],
+            [is_finalizer_message],
+            [],
+        ),
+    ],
```
muralov, line 89: can we also add test case where all three filters are considered? 
`[is_human_message, is_finalizer_message, is_system_message]`

src/agents/supervisor/agent.py
```diff
@@ -130,9 +137,15 @@ def _create_planner_chain(self, model: IModel) -> RunnableSequence:
 
     def _invoke_planner(self, state: SupervisorState) -> AIMessage:
         """Invoke the planner."""
+
+        messages = filter_messages_via_checks(
```
muralov, line 141: are we not ignoring such following messages. I found them in the Langfuse traces. 
The information about resource is quite important.  I don't why it is not shown as system messages. 
Can you check whether if your langfuse traces still have them?
```
  AIMessage(
      content="The user query is related to: {'resource_api_version': 'gateway.kyma-project.io/v1beta1', 'resource_namespace': 'kyma-app-apirule-broken'}"
  ),
```

muralov (commented):
We need to adapt the integration tests too. For example, `test_kyma.py` integration tests use messages not filtered yet. Please take a look.

src/agents/supervisor/agent.py
friedrichwilken: we have system messages with the same content
```
SystemMessage(content="The user query is related to: {'resource_api_version': 'v1', 'resource_namespace': 'nginx-wrong-image'}", id='454913b7-4594-421d-828f-211eb841bfa4')
```

tests/unit/utils/test_filter_messages.py
friedrichwilken: sure

muralov (approved):

Issue #200, opened 2024-10-02
Implement initial version Retrieval-Augmented Generation (RAG) System for Kyma Agent


Teneroy:
## Description

The goal of this task is to implement a Retrieval-Augmented Generation (RAG) system for the Kyma agent. The RAG system will enhance the Kyma agent's ability to provide accurate and relevant responses by retrieving relevant documents from a knowledge base and combining the retrieved information with a generative model. The system should include proper query-translation techniques, efficient information retrieval, and a reranking mechanism to improve the quality of retrieved information. 

Ensure that the flow works, necessary steps are implemented. Further optimization of performance will go in the following tickets

## Subtasks

1. **Query Translation:**
   - Implement query-translation techniques to transform user queries into an optimal format for the retrieval system.
   - Ensure that the Kyma agent can handle diverse and complex queries by translating them into effective search queries.
   - Test different query-translation strategies to enhance the accuracy and relevance of the retrieval process.

2. **Information Retrieval:**
   - Develop a retrieval mechanism to pull relevant documents or data from the knowledge base, such as Kyma documentation or related resources.
   - Integrate the Hana Vector Database client as the main source for retrieving embedded information.

3. **Reranking Mechanism:**
   - Implement a reranking system that scores and ranks the retrieved documents based on their relevance to the original query.
   - Use relevant metrics or machine learning-based approaches to ensure that the most important and accurate information is ranked higher.
   - Test the reranking system with a variety of test cases to validate its effectiveness.

## Acceptance Criteria

- [ ] Query-translation techniques are implemented and improve the quality of search queries - Mansur
- [ ] The information retrieval system pulls relevant documents from the knowledge base efficiently 
- [ ] A reranking mechanism is implemented and significantly improves the relevance of retrieved documents.

PR #296, opened 2024-11-27
feat: add initial Kyma RAG reranker

marcobebway:
**Description**

Implement Kyma RAG reranker. It reranks document list fetched from vector-db to answer the user queries (original query + generated queries). It uses the RRF algorithm as a first filtration layer to filter out the less relevant documents to the user queries. Then uses an LLM to rerank the remaining documents with regard to answering the user queries.

**Hints for reviewers**

- The deepeval thresholds for the RAG integration tests were lowered as requested by @muralov as he is already working on improving that on another ticket.
- The graph step timeout is increased for now to mitigate timeouts, later this should be improved and reduced again.

**Related issue(s)**

- #200


src/rag/reranker/reranker.py
```diff
@@ -0,0 +1,109 @@
...
+        :param output_limit: The maximum number of documents to return.
+        :return: A list of reranked documents.
+        """
+        logger.info(f"Reranking documents for queries: {queries}")
+
+        # filtration to prevent reranking irrelevant documents
+        relevant_docs = get_relevant_documents(docs_list, limit=input_limit)
+
+        # reranking using the LLM model
+        response: RerankedDocs = self.chain.invoke(
```
muralov, line 78:
```suggestion
        response: RerankedDocs = await self.chain.ainvoke(
```

src/rag/reranker/reranker.py
```diff
@@ -0,0 +1,109 @@
...
+    return "[{}]".format(",".join(document_to_str(doc) for doc in docs))
+
+
+def format_queries(queries: list[str]) -> str:
+    """
+    Format the queries for the prompt.
+    :param queries: A list of queries.
+    :return: A string representation of the queries in JSON format.
+    """
+    return "[{}]".format(",".join(f'"{query}"' for query in queries))
```
muralov, line 109: nitpick: if you have time it'd better to use Pydantic model here:
```
def format_documents(docs: List[Document]) -> str:
    """
    Format documents using Pydantic models.
    """
    reranked_docs = RerankedDocs(
        documents=[RerankedDoc(page_content=doc.page_content) for doc in docs]
    )
    return reranked_docs.model_dump_json()

def format_queries(queries: List[str]) -> str:
    """
    Format queries using Pydantic model.
    """
    class Queries(BaseModel):
        items: List[str]
    
    query_model = Queries(items=queries)
    return query_model.model_dump_json()
```

src/rag/reranker/prompt.py
```diff
@@ -0,0 +1,31 @@
...
+
+<input-queries>
+{queries}
+</input-queries>
+
+<your-tasks>
+1. Carefully read the queries and documents provided in the input sections.
+2. Compare the queries to each document by considering keywords or semantic meaning.
+3. Determine a relevance score for each document with respect to the queries (higher score means more relevant).
+4. Avoid documents that are irrelevant to the queries with very low relevance scores.
```
muralov, line 18: maybe you can try this instruction:
`4. Filter out documents that have low or no relevance to the queries based on their relevance scores.`
I am not sure if it improves, but you can give a try. If nothing changes, you can keep yours.

src/rag/reranker/reranker.py
muralov: please add error handling for reranking LLM call,  so that it fallbacks. It seems sometimes it cannot return the expected format and fails with parsing. This happened for mytesting. For example:
```
        try:
            response: RerankedDocs = await self.chain.ainvoke(
                {
                    "documents": format_documents(relevant_docs),
                    "queries": format_queries(queries),
                    "limit": output_limit,
                }
            )
        except Exception as e:
            logger.error(f"Error during document reranking: {str(e)}")
            # Fallback to using the filtered documents
            return relevant_docs[:output_limit]
```

src/rag/reranker/rrf.py
muralov, line 1: I like this one :)

src/rag/system.py
```diff
@@ -89,19 +80,17 @@ async def aretrieve(self, query: Query, top_k: int = 5) -> list[Document]:
...
         )
-        # remove duplicates from all retrieved documents
-        all_docs = self._remove_duplicates([doc for docs in docs_lists for doc in docs])
-
-        # TODO: re-rank documents
 
-        if len(all_docs) > top_k:
-            all_docs = all_docs[:top_k]
+        # rerank documents
+        reranked_docs = self.reranker.rerank(
```
muralov, line 88:
```suggestion
        reranked_docs = await self.reranker.arerank(
```

src/rag/reranker/reranker.py
muralov: we really need to test prompt of Reranker with separate integration tests. You can create a separate method for calling the chain only. Then it'll be really easy to test your reranking logic separately: 
```
def _anvoke_reranker():
  self.chain.invoke(...)
```

If you say too much for the PR, let's do it in a follow-up PR.

muralov (changes requested):
It looks really good. 
Follow-ups:
- fix the integration tests
- add more separate integration tests for re-ranking only

src/rag/reranker/reranker.py
```diff
@@ -74,21 +75,30 @@ def rerank(
...
+            response: RerankedDocs = await run_in_executor(
+                None,
+                self.chain.invoke,
+                {
+                    "documents": format_documents(relevant_docs),
+                    "queries": format_queries(queries),
+                    "limit": output_limit,
+                },
+                None,
+            )
```
muralov, line 89: can you just not use the following instead of run_in_executor?
```
response: RerankedDocs = await self.chain.ainvoke(
            {
                "documents": format_documents(relevant_docs),
                "queries": format_queries(queries),
                "limit": output_limit,
            }
        )
```

muralov (approved):

muralov (approved):

IwonaLanger (approved):

Issue #276, opened 2024-11-27
Develop response conversion node


Teneroy:
We need to develop a mechanism to convert any output which is being returned by the Finalizer into the format which would be agreed with Busola team. Apart from providing conveniently parsable response for busola team, we do it to allow an ability to navigate to and prefill yaml editor when we click "Setup in Editor"

**Acceptance criteria:**

- [x] Agree on a format of a response with busola team
- [x] Responses of companion returned by Finalizer are being converted to agreed format with busola team

PR #318, opened 2024-12-23
feat: added response conversion node

tanweersalah:
**Description**

- Added response conversion functionality to finalizer

**Related issue(s)**

#276


src/agents/common/response_converter.py
```diff
@@ -0,0 +1,262 @@
...
+                parsed_yaml = yaml.safe_load(yaml_config)
+
+        except Exception as e:
+            logger.error(
+                f"Error while parsing the yaml : {yaml_config} , Exception : {e}"
+            )
+
+            return None
+
+        return parsed_yaml
```
muralov, line 85: let's use TypedDict or Pydantic to parse:
```
class YAMLMetadata(TypedDict):
    namespace: str
    name: str

class YAMLConfig(TypedDict):
    metadata: YAMLMetadata
    kind: str

def _parse_yamls(self, yaml_config: str) -> Optional[YAMLConfig]:
    """
    Parse YAML string into Python object with error handling.
    Returns None if parsing fails.

    Args:
        yaml_config: YAML configuration string

    Returns:
        Parsed YAML object or None if parsing fails
    
    Example:
        >>> result = converter._parse_yamls("```yaml\\nkind: Deployment\\n```")
        >>> if result:
        >>>     print(f"Parsed YAML: {result}")
        >>> else:
        >>>     print("Failed to parse YAML")
    """
    try:
        if yaml_config[:7] == "```yaml":
            parsed_yaml = yaml.safe_load(yaml_config[8:-4])
        else:
            parsed_yaml = yaml.safe_load(yaml_config)
        
        return parsed_yaml

    except Exception as e:
        logger.error(
            f"Error while parsing the yaml : {yaml_config} , Exception : {e}"
        )
        return None
   ```

Comment (muralov) [reply src/agents/common/response_converter.py:85]: As I see if a yaml is generated but doesn't fit the schema, it returns None. 
Have you observed the cases where partial yaml resources are generated? We can discuss this.

File: src/agents/common/response_converter.py [hunk 2]
```diff
@@ -0,0 +1,262 @@
...
+    NEW_YAML,
+    NEXT,
+    UPDATE_YAML,
+)
+from utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+class ResponseConverter:
```
muralov, line 20: please create an interface IResponseConverter. That is our convention.

muralov (changes requested):
Please rebase on main, there is conflict.

I'd add such instructions for prompts so that LLM outputs fit your logic:
```
  1. For new resource deployments, wrap the YAML in <YAML-NEW> </YAML-NEW> tags
  2. For modifications to existing resources, wrap the YAML in <YAML-UPDATE> </YAML-UPDATE> tags
  3. Each YAML configuration must include:
     - apiVersion
     - kind
     - metadata (with at least name and namespace)
  4. Multiple YAML configurations should be separated into individual blocks with appropriate tags
  
    Example:
  <YAML-NEW>
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: my-new-app
    namespace: default
  spec:
    ...
  </YAML-NEW>

  <YAML-UPDATE>
  apiVersion: v1
  kind: Service
  metadata:
    name: existing-service
    namespace: production
  spec:
    ...
  </YAML-UPDATE>
 ```

src/agents/supervisor/agent.py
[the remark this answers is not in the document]
tanweersalah: i dont see either of them , 

https://langfuse.mur-nats.kymatunas.shoot.canary.k8s-hana.ondemand.com/project/cm4ty39mm000bifl0he3e61er/traces/1b0fd170-809b-47aa-82b7-b233a06db29b?timestamp=2025-01-23T15%3A34%3A09.422Z&observation=ca1b5345-a54d-4296-a05b-429fee764a86

muralov (approved):

Issue #256, opened 2024-11-13
Implement rate limiting


Teneroy:
Reasoning and Description:
In order to prevent high spendings and hitting AI core rate limiting, we need to introduce our own rate limiting per cluster which uses our companion. 

Tasks:
1. Make sure our traces in Langfuse have identifiers of clusters which uses our companion
Whenever we receive a request we need to:
2. Pull number of tokens used by cluster from 00:00UTC to 23:59UTC (possible through Langfuse API) or within 24h period(whatever is easier). 
3. If the number of tokens is higher than a constant we set, return a message to the user that they over consumed their token usage and should come back after 23:59UTC or after X minutes(if we go with 24h approach)

Acceptance criteria:

- [x] Traces are being identified per cluster
- [x] Total token usage is being pulled within agreed time range and compared against the constant

Teneroy:
https://api.reference.langfuse.com/#get-/api/public/observations

PR #331, opened 2025-01-09
feat: added rate limiter

tanweersalah:
**Description**

- Added langfuse api to get token usage
- Added rate limitation based on token usage
- Added unit test and integration test cases for langfuse api and rate limiter

**Related issue(s)**

#256


src/routers/conversations.py
```diff
@@ -147,3 +163,60 @@ async def messages(
...
+                                Langfuse API to retrieve token usage data.
+        token_limit : Default TOKEN_LIMIT_PER_CLUSTER
+
+    Raises:
+        HTTPException:  If the total token usage exceeds the daily limit (`TOKEN_LIMIT_PER_CLUSTER`),
+                        an HTTP 429 error is raised
+                        with details about the current usage,
+                        the limit, and the time remaining until the limit resets at midnight UTC.
+
+    """
```
friedrichwilken, line 189: Why is the style of the docstring here so different from the `get_daily_metrics` in `src/services/langfuse.py`?

tanweersalah: doc string were LLM generated, i have fixed the format now

tests/integration/services/test_langfuse.py
```diff
@@ -0,0 +1,157 @@
...
+    input_token_count,
+    companion_graph,
+    semantic_similarity_metric,
+    langfuse_service,
+):
+    # Conversation state with messages
+    state = create_mock_state(messages)
+
+    # Generate a random UUID
+    random_guid = uuid.uuid4()
```
friedrichwilken, line 121: is the `g` in `random_guid` a typo?

friedrichwilken: Oh btw the uuid has a property called `hex` that is basically what you want, so you would not need to remove the dashes.
```
my_id = uuid.uuid4().hex
```
but it is not important

tests/integration/services/test_langfuse.py
```diff
@@ -0,0 +1,157 @@
...
+@pytest.mark.asyncio
+async def test_langfuse_end_to_end(
+    messages,
+    input_token_count,
+    companion_graph,
+    semantic_similarity_metric,
+    langfuse_service,
+):
+    # Conversation state with messages
+    state = create_mock_state(messages)
```
friedrichwilken, line 118: Can you add a little doc string here that roughly outlines what the test does how it archives that. Something like
```
"""
The purpose of this test is to test an end to end scenario for langfuse by providing x and y, doing this and that, and  then assessing that something happend.
"""
```
It makes the life of people who have to deal with failing tests a little easier.

tests/unit/services/test_langfuse.py
```diff
@@ -0,0 +1,295 @@
...
+            "user123",
+            MetricsResponse(**mock_metrics_data),
+        ),
+    ],
+)
+@pytest.mark.asyncio
+async def test_get_daily_metrics(
+    from_timestamp, to_timestamp, tags, user_id, expected_result
+):
+    with patch("aiohttp.ClientSession.get") as mock_get:
```
friedrichwilken, line 116: Can you add a little doc string here that roughly outlines what the test does how it archives that. Something like
```
The purpose of this test is to test an ... by providing x and y, doing this and that, and  then assessing that something happend.
```
It makes the life of people who have to deal with failing tests a little easier.

friedrichwilken (changes requested):

tests/integration/services/test_langfuse.py
tanweersalah: @friedrichwilken , GUID (aka UUID) is an acronym for 'Globally Unique Identifier'

tests/unit/services/test_langfuse.py
tanweersalah: added docstring

tests/integration/services/test_langfuse.py
tanweersalah: added docstring

tests/integration/services/test_langfuse.py
tanweersalah: added docstring

PR #341, opened 2025-01-16
feat: add user, password and database-number as configurables to build redis URL

friedrichwilken:
…s URL

**Description**

- add user, password and database-number as configurables to build redis URL

**Related issue(s)**

### DIFF
--- a/src/utils/settings.py
+++ b/src/utils/settings.py
@@ -69,11 +69,11 @@ LOG_LEVEL = config("LOG_LEVEL", default="INFO")
 # A Redis URL has the format "redis://<username>:<password>@<host>:<port>/<db_number>
 REDIS_HOST = config("REDIS_HOST", default="localhost")
 REDIS_PORT = config("REDIS_PORT", default=6379, cast=int)
-REDIS_USER = config("REDIS_USER", default="")
-REDIS_PASSWORD = config("REDIS_PASSWORD", default="")
-REDIS_DB_NUMBER = config("REDIS_DB_NUMBER", default=0, cast=int)
-user_part = f"{REDIS_USER}:" if REDIS_USER else ""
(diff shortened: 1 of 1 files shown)


src/utils/settings.py
```diff
@@ -66,9 +66,15 @@ def load_env_from_json() -> None:
...
 # Redis
+# A Redis URL has the format "redis://<username>:<password>@<host>:<port>/<db_number>
 REDIS_HOST = config("REDIS_HOST", default="localhost")
 REDIS_PORT = config("REDIS_PORT", default=6379, cast=int)
-REDIS_URL = f"redis://{REDIS_HOST}:{REDIS_PORT}/0"
+REDIS_USER = config("REDIS_USER", default="")  # optional
+REDIS_PASSWORD = config("REDIS_PASSWORD", default="")  # optional
+REDIS_DB_NUMBER = config("REDIS_DB_NUMBER", default=0, cast=int)  # optional
+user_part = f"{REDIS_USER}" if REDIS_USER else ""
+auth_part = f"{user_part}:{REDIS_PASSWORD}@" if REDIS_USER or REDIS_PASSWORD else ""
```
mfaizanse, line 76: Should be a AND operator or OR?

friedrichwilken: It should be an "or" in my opinion

in theory you could log in without a username --users are rather new to redis:
```
redis://:mypw@localhost:67379/0
```
but you could also log in via a username when there is no pw required (which is an edge case but since it is so easy to do I though, why rule it out?)
```
redis://myuser:@localhost:67379/0
```
if you do not a pw or a user you just leave it all out of the url, including the ':' and '@'
```
redis://localhost:67379/0
```

If you think username + no pw is too much of an edge case, I can remove it. What do you think?

Issue #343, opened 2025-01-20
Improve Secret Management and Test Configuration Handling


Teneroy:
**Secret Management**
Currently, it is really hard to update GH Action secrets as GH doesn't display the values. Hence, we need to have the same names in GH and our storage. We should never forget to update our storage if GH secret is updated. 
The current existing unused secrets should be deleted too.

**Improve test configuration**
Currently, test configuration can be done in the `.env.test` which is old approach. Normally all configuration is done in `config.json` file. 
We should set different config path in .env file, e.g. `CONFIG_PATH=config/config.test.json`, if we want to configure for tests in a separate file.

mfaizanse:
- [ ] Update integration tests in pre-main.
- [ ] Update integration in release pipeline
- [ ] Validation tests
- [ ] Evaluation tests

PR #348, opened 2025-01-22
chore: refactor configs

mfaizanse:
**Description**

- Remove support for `.env.test` file.

**Related issue(s)**

- https://github.com/kyma-project/kyma-companion/issues/343


friedrichwilken (approved):

tests/blackbox/src/validation/utils/models.py
```diff
@@ -1,21 +1,29 @@
...
 from gen_ai_hub.proxy.native.google_vertexai.clients import GenerativeModel
 from pydantic import BaseModel
 
 logger = get_logger(__name__)
 
-proxy_client = get_proxy_client("gen-ai-hub")
+proxy_client: BaseProxyClient | None = None
+
+
+def get_gen_ai_proxy_client() -> BaseProxyClient:
```
muralov, line 16: nitpick: you can use `lru_cache` if you want to avoid global variable.

tests/blackbox/src/common/config.py
muralov, line 1: just a comment, you don't have to do it now. 

We need to share the model logic from a shared package. There is an issue already we'll do it later.

PR #350, opened 2025-01-22
fix: update enhance query generation error handling

muralov:
**Description**:
Improves error handling for multi query generation.

**Changes**:
- Modify error handling in QueryGenerator to log the original query when an error occurs during query generation, ensuring fallback to the original query.
- Update unit tests to reflect changes in error handling and improve clarity in parameter names.


tests/unit/rag/test_query_generator.py
```diff
@@ -80,29 +80,25 @@ def test_create_chain(self, mock_model, mock_llm):
...
+                mock_chain.ainvoke.side_effect = error
             else:
                 mock_chain.ainvoke.return_value = chain_output
 
-                result = await generator.agenerate_queries(query)
-                assert result == expected_output
-                mock_chain.ainvoke.assert_called_once_with({"query": query})
+            result = await generator.agenerate_queries(query)
+            assert result == expected_output
+            mock_chain.ainvoke.assert_called_once_with({"query": query})
```
friedrichwilken, line 104:
```suggestion
            # When
            result = await generator.agenerate_queries(query)
            # Then
            assert result == expected_output
            mock_chain.ainvoke.assert_called_once_with({"query": query})
```

tests/unit/rag/test_query_generator.py
```diff
@@ -80,29 +80,25 @@ def test_create_chain(self, mock_model, mock_llm):
...
     @pytest.mark.asyncio
     async def test_agenerate_queries(
-        self, mock_model, mock_llm, query, chain_output, expected_output, expected_error
+        self, mock_model, mock_llm, query, chain_output, expected_output, error
     ):
         """Test agenerate_queries method."""
         mock_chain = AsyncMock()
         with patch.object(QueryGenerator, "_create_chain", return_value=mock_chain):
             generator = QueryGenerator(mock_model)
             # When/Then
```
friedrichwilken, line 96:
```suggestion
        # Given
        mock_chain = AsyncMock()
        with patch.object(QueryGenerator, "_create_chain", return_value=mock_chain):
            generator = QueryGenerator(mock_model)
```

tests/unit/rag/test_query_generator.py
```diff
@@ -80,29 +80,25 @@ def test_create_chain(self, mock_model, mock_llm):
...
-            if expected_error:
-                mock_chain.ainvoke.side_effect = expected_error
-
-                with pytest.raises(type(expected_error)) as exc_info:
-                    await generator.agenerate_queries(query)
-                assert str(exc_info.value) == str(expected_error)
+            if error:
+                mock_chain.ainvoke.side_effect = error
             else:
                 mock_chain.ainvoke.return_value = chain_output
```
friedrichwilken, line 100: Could this now conceal errors that we might want to detect?

friedrichwilken (changes requested):

tests/unit/rag/test_query_generator.py
muralov: yes, it writes the error in the log with stacktrace. However, it still returns the original query as a fallback so that RAG can still search for documents with the original query.

friedrichwilken: If you say yes, the changed test will conceal errors will that we actually would want to detect here, why would we do this?

friedrichwilken (changes requested):

friedrichwilken (approved):

Issue #335, opened 2025-01-13
Shorten intermediate outputs of the companion


Teneroy:
### Description
In our UI we won't be able to display full intermediate answers of our agents. Therefore, we would need to display smaller intermediate steps. We can use subtasks for that and simply display subtasks for the beginning. Additionally, if you find it useful and simple to implement, we can output messages we are passing to the tools, so that users will be aware which resources are being retrieved and then analyzed.

### Example:

`{"event": "agent_action", "data": {"agent": "Supervisor", "answer": {"content": "Give me the list of namespaces again", "next": "KubernetesAgent"}}}`

convert to:

` {"event": "agent_action", "data": {"agent": "Supervisor", "answer": {"content": "Supervisor agent created a plan", "next": "KubernetesAgent"}}}`

`{"event": "agent_action", "data": {"agent": "KubernetesAgent", "answer": {"content": "Here is the list of namespaces in your Kubernetes cluster:\n\n1. bitnami-role-missing\n2. bitnami-wrong-rbac-permissions\n3. busybox-no-kubectl-binary\n4. cert-manager\n5. default\n6. improper-persistent-storage\n7. istio-system\n8. k8s-incorrect-liveness-probe\n9. k8s-wrong-svc-labels\n10. kcp-system\n11. kube-node-lease\n12. kube-public\n13. kube-system\n14. kyma-app-apirule-broken\n15. kyma-app-serverless-syntax-err\n16. kyma-eventing-wrong-event-type\n17. kyma-serverless-function-no-replicas\n18. kyma-subscription-old-event-type\n19. kyma-system\n20. nginx-oom\n21. nginx-wrong-image\n22. whoami-too-many-replicas\n\nIf you need more information about any specific namespace, feel free to ask!"}}}`

convert to:

 `{"event": "agent_action", "data": {"agent": "KubernetesAgent", "answer": {"content": "Kyma Agent retrieved namespaces"}}}`

final output:

[...]

PR #358, opened 2025-01-27
feat: shorten intermediate outputs of the companion

tanweersalah:
**Description**

- Removed Summarize Node from output chunk
- Made all the output chunks consistent
- Added subtask with a shorten steps description which can be shown to the user in frontend

**Related issue(s)**

#335


src/agents/common/state.py
```diff
@@ -28,9 +28,12 @@ class SubTask(BaseModel):
     description: str = Field(
         description="user query with original wording for the assigned agent"
     )
+    task_title: str = Field(
+        description="""Generate a title for 4 to 5 words, only use these:
```
friedrichwilken, line 32:
```suggestion
        description="""Generate a title of 4 to 5 words, only use these:
```
Just to double check: You try to say something like "Generate a title that consists out of 4 to 5 words", right?

src/agents/common/state.py
```diff
@@ -28,9 +28,12 @@ class SubTask(BaseModel):
...
+          Retrieving, Fetching, Extracting or Checking. Never user creating .."""
```
friedrichwilken, line 33:
```suggestion
          'Retrieving', 'Fetching', 'Extracting' or 'Checking'. Never use 'Creating'."""
```

friedrichwilken (changes requested):

friedrichwilken (approved):

PR #367, opened 2025-02-05
feat: start options to support debugpy

friedrichwilken:
**Description**

- add start options in `main.py` to allow debugg via [`debugpy`](https://pypi.org/project/debugpy/), the DAP implementation for Python that is supportet by NVim, as mentioned in the [official docs of FastAPI](https://fastapi.tiangolo.com/tutorial/debugging/)

**Related issue(s)**

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -28,9 +28,4 @@ async def healthz() -> dict:
 
 
 if __name__ == "__main__":
-    uvicorn.run(
-        "main:app",
-        host="localhost",
-        port=8000,
-        reload=True,
-    )
+    uvicorn.run(app, host="localhost", port=8000)
--- a/src/main.py
+++ b/src/main.py
@@ -1,3 +1,4 @@
+import uvicorn
 from fastapi import FastAPI
 
 from routers import conversations
@@ -24,3 +25,7 @@ async def readyz() -> dict:
 async def healthz() -> dict:
     """The endpoint for the Health Probe."""
     return {"healthy": "true"}
+
+
+if __name__ == "__main__":
+    uvicorn.run(app, host="localhost", port=8000)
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -24,3 +25,7 @@ async def readyz() -> dict:
 async def healthz() -> dict:
     """The endpoint for the Health Probe."""
     return {"healthy": "true"}
+
+
+if __name__ == "__main__":
+    uvicorn.run(app, host="localhost", port=8000)
```
muralov, line 31: do you need this to run the app with `python src/main.py`?

friedrichwilken: yes, if I want to debug it, I do need this, as the referenced docs of FastAPI shows.

muralov: Somehow `fastapi` cli is the recommended way in the docs https://fastapi.tiangolo.com/#standard-dependencies. 
Can you share the doc that you read? I really used the way and there were some issue for me.

friedrichwilken: https://fastapi.tiangolo.com/tutorial/debugging/

src/main.py
```diff
@@ -24,3 +25,12 @@ async def readyz() -> dict:
...
     return {"healthy": "true"}
+
+
+if __name__ == "__main__":
+    uvicorn.run(
+        "main:app",
+        host="localhost",
+        port=8000,
+        reload=True,
+    )
```
muralov, line 36: Actually your first change was correct, we don't need reload because we can provide `--reload` with IDEs (PyCharm, VS Code). The reason why we disable reload is that prod should have have auto-reload. 
just small improvements. Can you apply the following? 
```suggestion
if __name__ == "__main__":
    uvicorn.run(
        app,
        host="0.0.0.0",
        port=8000,
    )
```

muralov: Please document that we should now provide auto reload with with IDEs. 
There is --reload argument.

muralov (approved):

Issue #283, opened 2024-11-27
Develop cleanup mechanism for conversation history


Teneroy:
We don't want to store the conversation history indefinitely. Let's clean the history which is older than 12 hours. 

Acceptance criteria:

- [ ] Conversation history which was stored for more than 12 hours is being cleaned

PR #369, opened 2025-02-06
feat: cleanup mechanism for conversation history

tanweersalah:
**Description**

- Added TTL for each data point

**Related issue(s)**

#283


muralov (changes requested):
good job, it looks good. 

Can you please add integration tests? You can use fakeredis for that. 
You can just assert in the test if a key is deleted after certain seconds, by faking REDIS_TTL

PR #390, opened 2025-02-19
Fix security issues for checkmarx

muralov:
**Description**
Fix checkmarx security issues and improve error logging

- Fix security issues
- Improve exceptioon handling and logging errros

**Related issue(s)**
#375

### DIFF
--- a/doc_indexer/src/fetcher/fetcher.py
+++ b/doc_indexer/src/fetcher/fetcher.py
@@ -52,11 +52,9 @@ class DocumentsFetcher:
         try:
             scroller = Scroller(repo_dir, module_output_dir, source)
             scroller.scroll()
-        except Exception as e:
-            logger.error(
-                f"Error while scrolling documents for: {source.name}: {str(e)}"
-            )
-            raise e
+        except Exception:
+            logger.exception(f"Error while scrolling documents for: {source.name}")
+            raise
         finally:
             # delete the directories if they exist
             logger.debug(f"Deleting the temporary directory: {repo_dir}")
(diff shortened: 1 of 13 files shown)


doc_indexer/src/fetcher/fetcher.py
```diff
@@ -52,11 +52,9 @@ def fetch_documents(self, source: DocumentsSource) -> None:
...
             scroller = Scroller(repo_dir, module_output_dir, source)
             scroller.scroll()
-        except Exception as e:
-            logger.error(
-                f"Error while scrolling documents for: {source.name}: {str(e)}"
-            )
-            raise e
+        except Exception:
+            logger.exception(f"Error while scrolling documents for: {source.name}")
+            raise
```
mfaizanse, line 57: Will `raise` implicitly raise the actual Exception?

mfaizanse: Also, shall we document somewhere these common recommendations to avoid future security issues?

muralov: yes, it re-raises the current exception.

I've just refactored it and the error is still logged in the error log. It is normal to log the error that we need in the logs. No end user can see it.

The issue was with the error messages added to messages that I removed.

mfaizanse: shall we document somewhere these common recommendations to avoid future security issues?

muralov: this is just refactoring that is not related to security issue.

mfaizanse (approved):

Issue #391, opened 2025-02-20
Only allow conversation owners to continue the chat


mfaizanse:
Currently, we do not have any authorization for the conversation endpoint. An attacker can get the conversation id for a different user and try to get the conversation history.

To mitigate the issue, 
- we will save owner information (get the owner information from the k8s token) of the conversation in Redis. 
- Then authorize the user in conversation endpoint, only allow to continue the conversation if the request is made by the owner.

**AC(s):**
- [ ] Users can only only interact with their own conversations.
- [ ] Appropriate HTTP Status Code is return if the authorization fails.

PR #402, opened 2025-02-26
chore: added authorization for conversation

mfaizanse:
**Description**

- added authorization for conversation

**Related issue(s)**

- https://github.com/kyma-project/kyma-companion/issues/391


src/utils/utils.py
```diff
@@ -40,3 +43,29 @@ def string_to_bool(value: str) -> bool:
...
+        if "sub" in payload and payload["sub"] != "":
+            return str(payload["sub"])
+        elif "email" in payload and payload["email"] != "":
+            return str(payload["email"])
+        elif (
+            "kubernetes.io/serviceaccount/service-account.name" in payload
+            and payload["kubernetes.io/serviceaccount/service-account.name"] != ""
+        ):
+            return str(payload["kubernetes.io/serviceaccount/service-account.name"])
+        raise Exception("Invalid token: User identifier not found in token")
```
muralov, line 69: nitpick
```suggestion
        raise ValueError("Invalid token: User identifier not found in token")
```

src/utils/utils.py
```diff
@@ -40,3 +43,29 @@ def string_to_bool(value: str) -> bool:
...
+
+def parse_k8s_token(token: str) -> Any:
+    """Decode the JWT token without verification."""
+    try:
+        return jwt.decode(token, options={"verify_signature": False})
+    except Exception as e:
+        raise ValueError("Failed to parse k8s token") from e
+
+
+def get_user_identifier_from_token(token: str) -> str:
```
muralov, line 56: it'd be nice if you explain what it does as a part of function doc. I am bit newer to the authentication types: sub, email, service-account

src/services/conversation.py
muralov, line 1: It'd be nice if we log the errors without exposing any sensitive token information. 
It'll help to identify the cause if error occurs.

nitpick: please use constants for the string hardcoded values. Otherwise, security tools (e.g. checkmarx) may confuse and create an issue.

tests/unit/utils/test_src_utils.py
```diff
@@ -51,3 +56,84 @@ def test_is_empty_str(input_data, expected_output):
...
+            {"email": "user@example.com"},
+            "user@example.com",
+            None,
+        ),
+        (
+            "valid token with service account name",
+            {"kubernetes.io/serviceaccount/service-account.name": "service-account"},
+            "service-account",
+            None,
+        ),
```
muralov, line 113: nitpick: you can re-use the constants SUB, EMAIL, SERVICE_ACCOUNT_NAME here to if you define them in the main code.

tests/unit/agents/test_graph.py
```diff
@@ -498,3 +498,106 @@ async def test_aget_messages(
...
+            ),
+            (
+                "conversation2",
+                "user2",
+                {},
+                {"thread_owner": "user2"},
+            ),
+        ],
+    )
+    async def test_aupdate_thread_owner(
```
muralov, line 569: nitpick: you could have used fake-redis too.

muralov (changes requested):
pretty good job. 
We may also use thread_owner in addition to thread ID in the async_redis and other places code to retrieve thread related data in the future.

src/utils/utils.py
mfaizanse: Its a [JWT token](https://jwt.io/), and the payload of a JWT token commonly contains `sub` as a unique identifier. Based on different [k8s authentication options](https://kubernetes.io/docs/reference/access-authn-authz/authentication/), I have used the keys to get the data. Also, I have verified a token from SKR and the token contains `sub` and `email`.

src/utils/utils.py
mfaizanse: updated

src/services/conversation.py
mfaizanse: added log line for error. also changed hard coded strings as constants.

tests/unit/utils/test_src_utils.py
mfaizanse: done

tests/unit/agents/test_graph.py
mfaizanse: IMO, `self.graph.aupdate_state` is a method from langchain method so mock at this level should be enough.

muralov (approved):

Issue #406, opened 2025-02-27
Refactor rate limiting feature to not use LangFuse


mfaizanse:
As part of this [ticket](https://github.com/kyma-project/kyma-companion/issues/256), we used the LandFuse APIs to get the information about token usage and thus implemented rate limiting based on this information.

Now, we want to decouple the dependency of LangFuse. Therefore, we need to implement the rate limiting using the Redis instead.

**AC(s):**
- [x] Custom logic is added to store token usage information per cluster in Redis. (Hint: check LangChain [callbacks](https://python.langchain.com/docs/concepts/callbacks/))
- [x] ACs of this [ticket](https://github.com/kyma-project/kyma-companion/issues/256) are still covered.

mfaizanse:
- [ ] Remove LangfuseService class as not used anymore

Issue #392, opened 2025-02-20
Connect readiness and liveness probes endpoints with actual health of companion


mfaizanse:
Connect readiness (`/readyz`) and liveness (`/healthz`) probes [endpoints](https://github.com/kyma-project/kyma-companion/blob/main/src/main.py) with actual health of companion.

**AC(s):**
- [ ] The readiness endpoint returns true only when the connections for the Redis, HanaDB and AICore are initialized
- [ ] The liveness probe return true only when the connections for the Redis, HanaDB and AICore are connected.
  - [ ] Also check if there are any failures in [UsageTrackerCallback](https://github.com/kyma-project/kyma-companion/blob/0bd7c6a58f8baa1792ab13cda91ecdc9125c9adc/src/services/usage.py#L42). If the callback is failing to push the data to redis, it will not fail the graph. So, we need to separately track its health.

mfaizanse:
- [ ] also update https://github.tools.sap/kyma/management-plane-charts/tree/chart/kyma-companion

PR #415, opened 2025-03-03
feat: added usage tracking for token limit based on Redis

mfaizanse:
**Description**

- added usage tracking for token limit based on Redis.
- added logic in blackbox tests to verify that token usage data is pushed to Redis.
  
   image

**Related issue(s)**

- https://github.com/kyma-project/kyma-companion/issues/406


src/services/usage.py
```diff
@@ -0,0 +1,201 @@
...
+        response: LLMResult,
+        *,
+        run_id: UUID,
+        parent_run_id: UUID | None = None,
+        **kwargs: Any,
+    ) -> Any:
+        """Overridden callback method to track the token usage."""
+        usage = _parse_usage(response)
+        if usage is None:
+            raise ValueError("Usage information not found in the LLM response.")
```
tanweersalah, line 42: will this stop the flow of complete graph ?

src/agents/memory/async_redis_checkpointer.py
```diff
@@ -435,3 +476,37 @@ async def _aget_checkpoint_key(
...
+    async def awrite_llm_usage(self, cluster_id: str, data: dict, ttl: int = 0) -> str:
+        """Write LLM usage data to Redis. Return the key."""
+        key = _make_llm_usage_key(cluster_id)
+        if ttl > 0:
+            await self.conn.set(key, json.dumps(data), ex=ttl)
+        else:
+            await self.conn.set(key, json.dumps(data))
+        return key
+
+    async def adelete_expired_llm_usage_records(
```
tanweersalah, line 489: why we need this , when we already added TTL ?

src/agents/memory/async_redis_checkpointer.py
```diff
@@ -435,3 +476,37 @@ async def _aget_checkpoint_key(
...
+        self, cluster_id: str, ttl: int
+    ) -> None:
+        """Delete expired LLM usage records.
+        Returns the list of keys that were deleted."""
```
tanweersalah, line 493: There is no Return statement, so this comment is not valid

src/agents/memory/async_redis_checkpointer.py
```diff
@@ -435,3 +476,37 @@ async def _aget_checkpoint_key(
...
+        keys_to_delete = []
+        for key in keys:
+            old_time = _extract_time_from_llm_usage_key(_safe_decode(key))
+            if time.time() - old_time > ttl:
+                keys_to_delete.append(key)
+        if len(keys_to_delete) > 0:
+            await self.conn.delete(*keys_to_delete)
+
+    async def alist_llm_usage_records(self, cluster_id: str, ttl: int) -> list[dict]:
+        """List non-expired LLM usage records."""
```
tanweersalah, line 504: maybe you can update the doc string with return statement

src/routers/conversations.py
```diff
@@ -195,58 +196,19 @@ async def messages(
...
             status_code=ERROR_RATE_LIMIT_CODE,
             detail={
                 "error": "Rate limit exceeded",
-                "message": f"Daily token limit of {token_limit} exceeded for this cluster",
-                "current_usage": total_token_usage,
-                "limit": token_limit,
-                "time_remaining_seconds": seconds_remaining,
+                "message": "Daily token limit exceeded for this cluster",
+                "limit": TOKEN_LIMIT_PER_CLUSTER,
+                "time_remaining_seconds": TOKEN_USAGE_RESET_INTERVAL,
```
tanweersalah, line 209: we want to return remaining time. if that is not possible then we can change the key to something "reset_interval" , in my implementation we consider to reset the token usage midnight UTC

src/routers/conversations.py
```diff
@@ -195,58 +196,19 @@ async def messages(
...
             },
-            headers={"Retry-After": str(seconds_remaining)},
+            headers={"Retry-After": str(TOKEN_USAGE_RESET_INTERVAL)},
```
tanweersalah, line 211: this also incorrect as it will always fixed

src/services/conversation.py
```diff
@@ -150,3 +166,10 @@ async def authorize_user(self, conversation_id: str, user_identifier: str) -> bo
             return True
         # If the owner is the same as the user, we can authorize the user.
         return owner == user_identifier
+
+    async def is_usage_limit_exceeded(self, cluster_id: str) -> bool:
+        """Check if the token usage limit is exceeded for the given cluster_id."""
+        # Delete expired records before checking the usage limit.
+        await self._usage_limiter.adelete_expired_records(cluster_id)
```
tanweersalah, line 173: if we are good with the REDIS TTL, we dont need this.

src/agents/memory/async_redis_checkpointer.py
mfaizanse: I have added also custom logic to clear records to be on the safe side. Redis TTL seem to DEL the expired record with random sampling, as described in section `How Redis expires keys` [here](https://redis.io/docs/latest/commands/expire/).

src/services/conversation.py
mfaizanse: I have added also custom logic to clear records to be on the safe side. Redis TTL seem to DEL the expired record with random sampling, as described in section `How Redis expires keys` [here](https://redis.io/docs/latest/commands/expire/).

src/services/usage.py
```diff
@@ -0,0 +1,201 @@
...
+    async def adelete_expired_records(self, cluster_id: str) -> None:
+        """Delete the expired records for the given cluster_id."""
+        await self.memory.adelete_expired_llm_usage_records(
+            cluster_id, self.reset_interval_sec
+        )
+
+    async def ais_usage_limit_exceeded(self, cluster_id: str) -> bool:
+        """Check if the token limit is exceeded for the given cluster_id."""
+        if self.token_limit == -1:
+            return False
```
tanweersalah, line 75: this is not needed as the default value is no more -1

src/services/usage.py
mfaizanse: no, errors in callback will not effect the graph invocation. I will add a [task](https://github.com/kyma-project/kyma-companion/issues/392) here to monitor health of this callback.

src/agents/memory/async_redis_checkpointer.py
mfaizanse: updated

src/routers/conversations.py
mfaizanse: I have updated the logic.

src/routers/conversations.py
mfaizanse: I have updated the logic.

tanweersalah (approved):

PR #441, opened 2025-03-11
refactor: remove unused utils

friedrichwilken:
**Description**

- remove unused functions

**Related issue(s)**

### DIFF
--- a/src/agents/common/state.py
+++ b/src/agents/common/state.py
@@ -127,12 +127,6 @@ class CompanionState(BaseModel):
             )
         return list(self.messages)
 
-    def all_tasks_completed(self) -> bool:
-        """Check if all the sub-tasks are completed."""
-        if self.subtasks is None:
-            return True
-        return all(task.status == SubTaskStatus.COMPLETED for task in self.subtasks)
-
     class Config:
         arbitrary_types_allowed = True
         fields = {"k8s_client": {"exclude": True}}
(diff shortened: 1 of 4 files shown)


src/agents/common/utils.py
```diff
@@ -48,24 +48,28 @@ def agent_node(
...
-                return {
-                    MESSAGES: [AIMessage(content=result["output"], name=name)],
-                }
-            except Exception as e:
-                logger.error(f"Error in agent {name}: {e}")
-                return {
-                    ERROR: str(e),
-                    NEXT: EXIT,
-                }
+    if state.subtasks is not None:
```
muralov, line 51: nitpick:
```suggestion
    if state.subtasks:
```

src/agents/common/state.py
```diff
@@ -129,6 +129,8 @@ def get_messages_including_summary(self) -> list[MessageLikeRepresentation]:
 
     def all_tasks_completed(self) -> bool:
```
muralov, line 130: it seems this method is not used anywhere. Can you remove it?

friedrichwilken: Even better. Done.

src/agents/common/utils.py
muralov, line 1: I've noticed the following functions of the module is not used anywhere: 
- create_agent
- agent_node (the one you improved)
- next_step
- get_current_day_timestamps_utc
- hash_url

It'd be cool, if you can just remove them. Then we would not need to improve agent_node either.

friedrichwilken: okay, how about I do this in a second PR?

friedrichwilken: Good finding. I will delete them.

Issue #440, opened 2025-03-11
Bug: Planner directly answers instead of creating subtask


muralov:
**Description**
Planner directly answers with the namespace name. 

**Expected result**:
It has to create a plan with a Kyma subtask

**Actual result**
Here is the langfuse [trace](https://cloud.langfuse.com/project/cm8320ls209o2ad067nq5cejg/traces/f399fbe3-8502-4a1b-b5ec-62bd3f460e06?timestamp=2025-03-11T05%3A53%3A59.526Z&observation=69e17771-c722-41ce-a13c-13eb8c033043) and the curl command:

```
❯ curl -N -s -H 'Accept: text/event-stream' \
 -H 'Content-Type: application/json' \
 -H "X-Cluster-Certificate-Authority-Data: $TEST_CLUSTER_CA_DATA" \
 -H "X-Cluster-Url: $TEST_CLUSTER_URL" \
 -H "X-K8s-Authorization: $TEST_CLUSTER_AUTH_TOKEN" \
 --request POST \
 --data '{"query":"why the function is not working?", "resource_kind":"", "resource_api_version":"", "resource_name":"", "namespace":"kyma-app-serverless-syntax-err"}' \
 http://localhost:8000/api/conversations/867263/messages | jq;
{
  "event": "agent_action",
  "data": {
    "agent": "Supervisor",
    "answer": {
      "content": "The function is not working due to a syntax error in the serverless configuration.",
      "tasks": [],
      "next": "__end__"
    }
  }
}
```

tanweersalah:
- [ ] remove direct response from planner
- [ ] use common node instead of direct response
- [ ] No finalizer when only one 1 common subtask
- [ ] Donot send sub task if only 1 subtask belong to common node

PR #455, opened 2025-03-19
fix: planner responding directly without creating subtasks

tanweersalah:
**Description**

- Removed direct response from planner
- Made planner less complex
- Refactor planner unit and integration tests
- Added Gatekeeper node
- Added unit and integration test for gatekeeper node.
- Update response.py to handle gatekeeper responses

**Related issue(s)**

#440


src/agents/common/state.py
```diff
@@ -52,6 +62,10 @@ def is_error(self) -> bool:
         """Check if the task is error status."""
         return self.status == SubTaskStatus.ERROR
 
+    def is_common_subtask(self) -> Any:
```
muralov, line 65: I see this is not used anywhere. Can you delerte it?

src/agents/graph.py
```diff
@@ -203,6 +214,84 @@ async def _common_node(self, state: CompanionState) -> dict[str, Any]:
...
+            if response.forward_query:
+                logger.debug("Gatekeeper node forwarding the query")
+                return {
+                    NEXT: SUPERVISOR,
+                    MESSAGES: [
+                        AIMessage(
+                            content="",
+                            name=GATEKEEPER,
+                        )
+                    ],
```
muralov, line 268: do you really need to explicityl add messages? We don't need any message from gatekeeper.

tests/integration/agents/test_gatekeeper_node.py
```diff
@@ -0,0 +1,271 @@
...
+    conversation_sample_2,
+    conversation_sample_5,
+    conversation_sample_6,
+)
+from integration.conftest import convert_dict_to_messages, create_mock_state
+
+
+# Correctness metric for not general queries that needs planning
+@pytest.fixture
+def gatekeeper_correctness_metric(evaluator_model):
```
muralov, line 17: can you improve your metric like `semantic_similarity_metric` integration/conftest.py? 
I am not sure if you can just re-use it as it is missing input.

muralov (dismissed):
good job. 
There are only minor issues. Please take a look.

src/agents/common/state.py
tanweersalah: Removed

tests/integration/agents/test_gatekeeper_node.py
tanweersalah: Updated

muralov (approved):

Issue #451, opened 2025-03-13
Bug: Fails for cluster overview queries due to token exceeding


muralov:
**Description**

If kyma companions is asked a question for cluster overview it is crashing as the cluster tokens are exceeding the LLM accepted limit. 

Here is the Langfuse [trace](https://langfuse.fzn-dev.kymatunas.shoot.canary.k8s-hana.ondemand.com/project/cm87a9ffl0006vs07e1gfq6n0/traces/6c4ac3d9-7a3b-4509-b8a4-3022dbe745be?timestamp=2025-03-13T15%3A47%3A38.001Z&observation=bd6b251b-3eba-40de-a555-e610067b3c34)

PR #474, opened 2025-03-31
fix: token explosion issue for cluster overview

tanweersalah:
**Description**

- Added new tool for cluster overview
- Added unit test and integration test

**Related issue(s)**

#451


tests/integration/agents/k8s/test_k8s_agent.py
```diff
@@ -0,0 +1,314 @@
...
+                    description="What is wrong with api rule?",
+                    task_title="What is wrong with api rule?",
+                    assigned_to="KymaAgent",
+                ),
+                k8s_client=Mock(spec_set=IK8sClient),  # noqa
+                is_last_step=False,
+                remaining_steps=AGENT_STEPS_NUMBER,
+            ),
+            "k8s_query_tool",
+        ),
```
muralov, line 90: please delete this as it is Kyma Agent related query.  ApiRule is Kyma resource.

tests/integration/agents/k8s/test_k8s_agent.py
```diff
@@ -0,0 +1,314 @@
...
+        # Test case for pod logs query
+        (
+            KubernetesAgentState(
+                agent_messages=[],
+                messages=[
+                    SystemMessage(
+                        content="The user query is related to: {'resource_api_version': 'v1', 'resource_namespace': 'default', 'resource_kind': 'Pod'}"
+                    ),
+                    HumanMessage(
+                        content="What deployments exist in the 'kube-system' namespace?"
```
muralov, line 130: what happens if the query is "What deployments exist"?  It should be the same effect as system message has the namespace and kind.

tests/integration/agents/k8s/test_k8s_agent.py
```diff
@@ -0,0 +1,314 @@
...
+        ),
+        # Test case for namespace overview
+        (
+            KubernetesAgentState(
+                agent_messages=[],
+                messages=[
+                    SystemMessage(
+                        content="The user query is related to: {'resource_namespace': 'production'}"
+                    ),
+                    HumanMessage(content="Show me cluster-level resource usage"),
```
muralov, line 187: your query is about cluster, but the system message has the namespace. This is interesting case we need to test with the running companion

tests/integration/agents/k8s/test_k8s_agent.py
```diff
@@ -0,0 +1,314 @@
...
+        ),
+        # Test case for specific resource query with full details
+        (
+            KubernetesAgentState(
+                agent_messages=[],
+                messages=[
+                    SystemMessage(
+                        content="The user query is related to: {'resource_api_version': 'apps/v1', 'resource_namespace': 'default', 'resource_kind': 'Deployment'}"
+                    ),
+                    HumanMessage(content="Show me the details of my-app deployment"),
```
muralov, line 215: I am interested what happens if we ask "Show me the details of resource". As system message has the resource kind and namespace, it should still check the deployment.
Can you test?

muralov (changes requested):

tests/.deepeval_telemetry.txt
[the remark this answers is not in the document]
tanweersalah: done

tests/integration/agents/k8s/test_k8s_agent.py
tanweersalah: Added the testcase

tests/integration/agents/k8s/test_k8s_agent.py
tanweersalah: This particular test case is not stable, some time uses query_tool some time user overview_tools, so i not added this.

muralov: can you create an issue for this? 
Context has default namespace, but query asks about `kube-system` namespace.

tests/integration/agents/k8s/test_k8s_agent.py
muralov: have you tested this one?

tests/integration/agents/k8s/test_k8s_agent.py
tanweersalah: Yes I will create an issue , we can later try to optimize it. Maybe using similar promting teching like gatekeeper help in stabilize it better.

muralov (approved):

Issue #486, opened 2025-04-03
Next queries fail after it failed once for the same conversation


muralov:
**Description**
If error occurs for a conversation, the next queries fail for the conversation too.

**Query langfuse traces**
- https://cloud.langfuse.com/project/cm8320ls209o2ad067nq5cejg/traces/96e1d1ed-bfa6-4613-9282-5336bb6216c6?timestamp=2025-04-03T12:50:26.240Z&display=details

- https://cloud.langfuse.com/project/cm8320ls209o2ad067nq5cejg/traces/bf6a8bfa-020b-4a8b-8592-fd1bba601ae9?timestamp=2025-04-03T12:53:33.527Z&display=details

- https://cloud.langfuse.com/project/cm8320ls209o2ad067nq5cejg/traces/f61ec689-5327-476b-bd08-79ff35c9a464?timestamp=2025-04-03T12:54:20.339Z&display=details

Error example to reproduce:
```
❯ curl -N -s -H 'Accept: text/event-stream' \
 -H 'Content-Type: application/json' \
 -H "X-Cluster-Certificate-Authority-Data: $TEST_CLUSTER_CA_DATA" \
 -H "X-Cluster-Url: $TEST_CLUSTER_URL" \
 -H "X-K8s-Authorization: $TEST_CLUSTER_AUTH_TOKEN" \
 --request POST \
 --data '{"query":"how to create an application with Kyma and analyse the application", "resource_kind":"", "resource_api_version":"", "resource_name":"", "namespace":"kyma-app-serverless-syntax-err"}' \
 http://localhost:8000/api/conversations/8672108/messages | jq;
{
  "event": "agent_action",
  "data": {
    "agent": "Supervisor",
    "answer": {
      "content": "",
      "tasks": [
        {
          "task_id": 0,
          "task_name": "Fetching application creation steps",
          "status": "pending",
          "agent": "KymaAgent"
        },
        {
          "task_id": 1,
          "task_name": "Retrieving application analysis methods",
          "status": "pending",
          "agent": "KymaAgent"
        }
      ],
      "next": "KymaAgent"
    },
    "error": null
  }
}
{
  "event": "agent_action",
  "data": {
    "agent": "KymaAgent",
    "answer": {
[...]

PR #500, opened 2025-04-16
fix: added filter to clean invalid AIMessage and ToolMessage sequence

mfaizanse:
**Description**

- added filter to clean invalid AIMessage and ToolMessage sequence

**Related issue(s)**

- https://github.com/kyma-project/kyma-companion/issues/486

### DIFF
--- a/src/agents/common/agent.py
+++ b/src/agents/common/agent.py
@@ -22,7 +22,7 @@ from agents.common.constants import (
     SUMMARIZATION,
 )
 from agents.common.state import BaseAgentState, SubTaskStatus
-from agents.common.utils import filter_messages, should_continue
+from agents.common.utils import filter_messages, filter_valid_messages, should_continue
 from agents.summarization.summarization import MessageSummarizer
 from utils.chain import ainvoke_chain
 from utils.logging import get_logger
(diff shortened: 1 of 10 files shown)


src/agents/common/utils.py
```diff
@@ -47,6 +51,44 @@ def filter_messages(
...
+            # if the sequence of AIMessage and ToolMessage(s) is completed, then append the completed sequence.
+            if is_valid_sequence:
+                # append the AIMessage and its corresponding ToolMessages.
+                filtered.append(message)
+                for j, _ in enumerate(message.tool_calls):
+                    filtered.append(messages[i + j + 1])
+        else:
+            # append other valid messages.
+            filtered.append(message)
+    return filtered
```
friedrichwilken, line 89: You could avoid all the nested `for` loops here with something like
```suggestion
filtered: list[BaseMessage] = []
    i = 0

    while i < len(messages):
        message = messages[i]

        if isinstance(message, ToolMessage):
            # ToolMessage should not be added directly.
            i += 1
            continue

        if isinstance(message, AIMessage) and message.tool_calls:
            # Check if the next messages are tool calls as requested by AIMessage.
            tool_call_count = len(message.tool_calls)
            next_messages = messages[i + 1 : i + 1 + tool_call_count]

            if len(next_messages) == tool_call_count and all(
                isinstance(next_msg, ToolMessage) for next_msg in next_messages
            ):
                # Append the AIMessage and its corresponding ToolMessages.
                filtered.append(message)
                filtered.extend(next_messages)
                i += tool_call_count  # Skip the tool messages

        else:
            # Append other valid messages.
            filtered.append(message)

        i += 1

    return filtered
```

src/agents/common/agent.py
```diff
@@ -141,16 +141,17 @@ def _subtask_selector_node(self, state: BaseAgentState) -> dict[str, Any]:
...
-            inputs[AGENT_MESSAGES] = filter_messages(state.messages)
+            input_messages = filter_messages(state.messages)
 
+        # invoke the chain.
         response = await ainvoke_chain(
             self.chain,
-            inputs,
+            {
+                AGENT_MESSAGES: filter_valid_messages(input_messages),
+                "query": state.my_task.description,
```
friedrichwilken, line 153: What about the case that `description` is `None`?
```
    my_task: SubTask | None = None
```

friedrichwilken (changes requested):

src/agents/common/utils.py
mfaizanse: updated

src/agents/common/agent.py
mfaizanse: The agent will only be invoked when the subtask is not None.

src/agents/common/utils.py
friedrichwilken: Nice, thanks. 👍🏻

PR #555, opened 2025-06-10
chore: update the token limit

friedrichwilken:
**Description**

- set the token limit per day and user to `550_000`

**Related issue(s)**


mfaizanse:
we will update the value in helm chart.

PR #562, opened 2025-06-11
feat: enhance Kyma agent prompts for too broad queries and delete managed fields from resources

muralov:
**Description**
Kyma agent cannot handle too broad quries like "check all Kyma resources" not so well. We need to restrict it to certain resources for the time being. 
Delete managed fields field in metadata of k8s and kyma resources as it is not need to analyse resource.

- Updated Kyma agent prompts to require clarification for broad queries about Kyma resources, improving user interaction.
- Added new test cases to validate the agent's response when more information is needed from the user.
- Implemented a method in DataSanitizer to remove managedFields from metadata, ensuring cleaner resource definitions.
- Enhanced unit tests for DataSanitizer to cover the removal of managedFields in various scenarios.

### DIFF
--- a/src/agents/kyma/prompts.py
+++ b/src/agents/kyma/prompts.py
@@ -7,25 +7,37 @@ KYMA_AGENT_INSTRUCTIONS = f"""
 a. Analyse user query, resource information, message history and tool responses
 b. Identify query intent based on the analysis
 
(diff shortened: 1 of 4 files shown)


tanweersalah (approved):
lgtm

PR #573, opened 2025-06-15
chore: eval test to release workflow

friedrichwilken:
**Description**

- ...
- ...
- ...

**Related issue(s)**


.github/workflows/create-release.yml
```diff
@@ -165,6 +163,21 @@ jobs:
           echo "${{ secrets.INTEGRATION_TEST_CONFIG }}" | base64 --decode | jq > $GITHUB_WORKSPACE/config/config.json
           poetry run poe test-integration
 
+  run-evaluation-tests:
+    ## **IMPORTANT**: If any changes are made to how to run the evaluation tests
+    ## in the release-workflow, make sure to update
+    ## in the run-integration-test.yaml workflow as well.
+    name: Run evaluation tests
+    env:
+      IMAGE_NAME: "europe-docker.pkg.dev/kyma-project/prod/kyma-companion" # without tag.
```
mfaizanse, line 172: this is without tag. At this stage in the pipeline, we have not build the image for the tag. 
So, I think we can use the ref/sha to fetch status of build job of that commit, and use that tag to test.

PR #583, opened 2025-06-20
feat: added validation for allowed cluster hosts

mfaizanse:
**Description**

- added validation for allowed cluster hosts

**Related issue(s)**

### DIFF
--- a/tests/unit/services/test_k8s.py
+++ b/tests/unit/services/test_k8s.py
@@ -367,6 +367,21 @@ class TestK8sAuthHeaders:
                 False,
                 ValueError,
             ),
+            (
+                "Domain that contains allowed domain as substring (should be blocked)",
+                "https://api.notexample.com",
+                ["example.com"],
+                False,
+                None,
+            ),
+            ("Exact domain match", "https://example.com", ["example.com"], True, None),
+            (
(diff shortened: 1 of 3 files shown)


tests/unit/services/test_k8s.py
```diff
@@ -264,16 +272,126 @@ def test_get_auth_type(
...
+                False,
+                ValueError,
+            ),
+            (
+                "Hostname is None (should raise ValueError)",
+                "http://",
+                ["example.com"],
+                False,
+                ValueError,
+            ),
```
muralov, line 369: can you add the following test cases too? The first one is failing. Can you fix it? 
```
            (
                "Domain that contains allowed domain as substring (should be blocked)",
                "https://api.notexample.com",
                ["example.com"],
                False,
                None,
            ),
            ("Exact domain match", "https://example.com", ["example.com"], True, None),
            (
                "Port in URL should be handled correctly",
                "https://api.example.com:8080",
                ["example.com"],
                True,
                None,
            ),
```

muralov (changes requested):
it looks mainly good. There was one issue as I commented.

muralov (approved):

Issue #491, opened 2025-04-10
support feedback collection for responses


mfaizanse:
The idea is that companion would be able to recognize when the user is giving feedback or an opinion on its capabilities, and as a result the companion should prompt the user to fill the survey.

The flow would be:
1. Whenever a new message from user is received, detect if it a query or a feedback.
2. If it is a feedback, then return some flag to the frontend. The frontend will then ask the user to fill a survey.

**AC(s):**
- [ ] Companion is able to reliably identify a feedback message or a query.
- [ ] Feedback detection is working without effecting the performance of companion.
- [ ] Agree with frontend team on how to communicate [this](https://github.com/kyma-project/busola/issues/3852) to them.
   - [ ] Hint: Return task name: `Feedback`

PR #596, opened 2025-06-30
feat: handle feedback

muralov:
**Description**
Return feedback flag to the UI so that they can display feedback UI part for the end users

- Implement the logic in a separate LLM call and return the is_feedback flag as a part of Gatekeeper response
- Add/Update the int and unit tests

**Related issue(s)**
#491

### DIFF
--- a/src/agents/graph.py
+++ b/src/agents/graph.py
@@ -23,7 +23,6 @@ from agents.common.agent import IAgent
 from agents.common.constants import (
     COMMON,
     CONTINUE,
-    FEEDBACK,
     GATEKEEPER,
     INITIAL_SUMMARIZATION,
     MESSAGES,
@@ -272,9 +271,7 @@ class CompanionGraph:
         response = await ainvoke_chain(
             self._feedback_chain,
             {
-                "messages": filter_valid_messages(
-                    state.get_messages_including_summary()
-                ),
+                "messages": [state.messages[-1]],  # last human message
             },
         )
         return response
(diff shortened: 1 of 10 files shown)


tests/integration/agents/test_gatekeeper_feedback.py
```diff
@@ -0,0 +1,271 @@
...
+        (
+            "user gives shorter explanation",
+            [
+                SystemMessage(content="The user query is related to: "),
+                HumanMessage(content="give me a shorter explanation"),
+            ],
+            False,
+        ),
+    ],
+)
```
mfaizanse, line 253: Can you add a test case where there is little more conversation history?

mfaizanse (commented):
You need to add nano model deployment id to evaluation and integration tests config file.

src/agents/graph.py
```diff
@@ -232,42 +255,72 @@ def _create_gatekeeper_chain(model: IModel) -> RunnableSequence:
...
+        return cast(GatekeeperResponse, response)
 
     async def _gatekeeper_node(self, state: CompanionState) -> dict[str, Any]:
         """Gatekeeper node to handle general and queries that can answered from conversation history."""
 
         try:
-            response = await self._invoke_gatekeeper_node(state)
+            feedback_response = await self._invoke_feedback_node(state)
+        except Exception:
+            logger.exception("Error in feedback node")
```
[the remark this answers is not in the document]
muralov: the `logger.exception` logs exception strack trace automatically. This is special function

tests/integration/agents/test_gatekeeper_feedback.py
muralov: currently, feedback is only considering the user query. It doesn't consider conversation history. 
That's why I thought, it doesn't bring anything as it only uses only latest human message. 
There is another reason, where nano model cannot handle too much data why I only consider only user query.

PR #600, opened 2025-06-30
refactor: extracted if-else from gatekeeper prompt to python

mfaizanse:
**Description**

- extracted if-else from gatekeeper prompt to python

**Related issue(s)**

### DIFF
--- a/src/agents/common/state.py
+++ b/src/agents/common/state.py
@@ -89,15 +89,15 @@ class GatekeeperResponse(BaseModel):
         str,
         Field(
             description="""
-            Classifies 'user intent' into the following categories:
-            - "Kyma": kyma related user intent
-            - "Kubernetes": kubernetes related user intent
-            - "Programming": Programming related user intent (NOT specific to Kyma/Kubernetes)
-            - "About You": user intent about you and your capabilities
+            Classifies 'user query or intent' into the following categories:
+            - "Kyma": User query related to kyma related user intent.
+            - "Kubernetes": User query related to kubernetes.
+            - "Programming": User query related to Programming but NOT specific to Kyma or Kubernetes.
+            - "About You": User query about you and your capabilities and expertise.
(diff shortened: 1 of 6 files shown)


src/agents/common/state.py
```diff
@@ -26,36 +26,130 @@ class SubTaskStatus(str, Enum):
...
+            - If previous context discussed "Kyma Function" and user asks "check it?"
+            - The intent should be identified as "Check Kyma Function"
+                
+            The conversation history provides essential context for resolving ambiguous references.
+            """,
+        ),
+    ]
+
+    category: Annotated[
+        str,
```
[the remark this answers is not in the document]
tanweersalah: we already have this implemented in PLAN class , and till now i have never seen llm provinding any other value which result in parsing, This actually make llm to stick to these category which make it more deterministic  ,  And this will also solve the issue where llm sometime use different wording for same category like instead of `programming` it say `python-programming`

src/agents/common/state.py
```diff
@@ -26,36 +26,130 @@ class SubTaskStatus(str, Enum):
...
             """,
         ),
     ]
+
     forward_query: Annotated[
         bool,
         Field(
             default=False,
-            description="Flag to indicate to forward queries related to Kyma or Kubernetes.",
+            description="Always False.",
```
muralov, line 119: why do you want to have forward_query as `false` always?

mfaizanse: We do not need this value from LLM, it will be false by default and in our if-else, we change it to true if required.

tanweersalah: i just google i found this , 

we can make forward_query a private field in this way llm will not fill this value and not sent to llm , 

`_ forward_query: bool = PrivateAttr(default=False)`

another option is to make this excluded , 

`Annotated[bool, Field(default=False, exclude=True)]`

can you test which works

PR #676, opened 2025-07-24
chore: upgrade Dockerfile base image and hdbcli package

muralov:
**Description**
Upgrade base image and packages to fix security issues.

- Upgrade base image garden linux
- Upgrade hdbcli package


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/gardenlinux/gardenlinux:1592.11 AS builder
+FROM ghcr.io/gardenlinux/gardenlinux:1877.1 AS builder
```
friedrichwilken, line 1: on `1877.1` `python3.12` is no longer available. the only available version on that image would be `python3.13`
#734 would handle this by also bumping the python version.

Issue #701, opened 2025-07-28
Include the companion response in Partial masking mode of LangFuse traces


mfaizanse:
Currently, in Partial masking mode for LangFuse we only include the user query.

We need to also include the final response of the user query, so that we can monitor performance of companion

PR #718, opened 2025-07-29
feat: added FILTERED masking mode in langfuse service

mfaizanse:
**Description**

- added FILTERED masking mode in langfuse service.
- It will redact the tool messages and sanitize all other messages.

**Related issue(s)**

- https://github.com/kyma-project/kyma-companion/issues/701

### DIFF
--- a/src/routers/conversations.py
+++ b/src/routers/conversations.py
@@ -190,9 +190,7 @@ async def followup_questions(
 
     # Authorize the user to access the conversation.
     user_identifier = extract_user_identifier(k8s_auth_headers)
-    await authorize_user(
-        str(conversation_id), user_identifier, conversation_service
-    )
+    await authorize_user(str(conversation_id), user_identifier, conversation_service)
 
     # Check rate limitation
     await check_token_usage(x_cluster_url, conversation_service)
--- a/tests/unit/routers/test_conversations.py
+++ b/tests/unit/routers/test_conversations.py
@@ -16,7 +16,8 @@ from routers.conversations import (
     authorize_user,
     check_token_usage,
     enforce_query_token_limit,
-    init_conversation_service, extract_user_identifier,
+    extract_user_identifier,
+    init_conversation_service,
 )
 from services.conversation import IService
 from services.k8s import IK8sClient, K8sAuthHeaders


src/agents/graph.py
```diff
@@ -483,7 +484,11 @@ async def astream(
                 self.handler,
                 UsageTrackerCallback(cluster_id, cast(IUsageMemory, self.memory)),
             ],
-            tags=[cluster_id],  # cluster_id as a tag for traceability and rate limiting
+            tags=[cluster_id],
+            metadata=get_langfuse_metadata(
+                message.user_identifier if message.user_identifier else "unknown",
```
friedrichwilken, line 489: nit and more of an preference thingy
can be simplified to
```suggestion
                message.user_identifier or "unknown",
```

src/routers/conversations.py
```diff
@@ -326,7 +328,7 @@ async def authorize_user(
     conversation_id: str,
     k8s_auth_headers: K8sAuthHeaders,
     conversation_service: IService,
-) -> None:
+) -> str:
     """Authorize the user to access the conversation."""
```
friedrichwilken, line 332: with the new return, I think this docstring needs an update.

friedrichwilken: Also, this might be viewed as a violation of the single responsibility principle. What will happen if this will cause an exception and not return a str?

src/services/langfuse.py
```diff
@@ -101,75 +97,73 @@ def masking_production_data(self, data: Any) -> Any:
...
-    async def get_daily_metrics(
-        self,
-        from_timestamp: str,
-        to_timestamp: str,
-        tags: Any = None,
-        user_id: Any = None,
-    ) -> MetricsResponse | None:
+        return REDACTED
+
+    def _masking_mode_filtered(self, data: Any) -> Any:  # noqa: C901
```
friedrichwilken, line 117:
```suggestion
    def _masking_mode_filtered(self, data: Any) -> Any:  # noqa: C901
```
The linter has a point about the complexity here. I assume it would be pretty hard to reduce the complexity, here?

friedrichwilken (dismissed):

src/agents/graph.py
mfaizanse: updated

src/services/langfuse.py
mfaizanse: yes

src/routers/conversations.py
```diff
@@ -355,6 +357,16 @@ async def authorize_user(
             "Unable to get user identifier from the provided Authorization headers.",
         )
 
+    return user_identifier
```
tanweersalah, line 360: why we need to stop the KC flow if user is not found ??

mfaizanse: To not allow unauthorized users to access companion.

tanweersalah (approved):
lgtm

PR #772, opened 2025-08-26
feat: index with backup

friedrichwilken:
**Description**

- when reindexing docs, create a backup and restore if an error occurs

**Related issue(s)**


doc_indexer/src/indexing/indexer.py
```diff
@@ -83,47 +88,180 @@ def _load_documents(self) -> list[Document]:
...
+
+        # Create a backup of the current main table (if exists).
+        logger.info(
+            f"Backing up current table {self.db.table_name} to {self.backup_table_name} if exists..."
+        )
+        self._copy_table(
+            self.db.table_name,
+            self.backup_table_name,
+            only_warn_if_table_inexistent=True,
+        )
```
muralov, line 245: what if copying the table fails? Does it stop proceeding here?

doc_indexer/src/indexing/indexer.py
```diff
@@ -83,47 +88,180 @@ def _load_documents(self) -> list[Document]:
...
+            logger.exception(
+                "Error during indexing. Attempting to restore from backup."
+            )
+            self._drop_table(self.db.table_name, only_warn_if_table_inexistent=True)
+            self._rename_table(
+                self.backup_table_name,
+                self.db.table_name,
+                only_warn_if_table_inexistent=True,
+            )
+            raise
```
muralov, line 267: you can delete backup table if it exists with finally block

doc_indexer/src/indexing/indexer.py
friedrichwilken: It depends: 
If the table we are trying to copy does not exist, we only will log a warning (`only_warn_if_table_inexistent=True`).
If anything else happens, the process stops here.

doc_indexer/src/indexing/indexer.py
friedrichwilken: Valid point.

PR #789, opened 2025-08-31
fix: index unit tests do no longer require config.json

friedrichwilken:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/.github/workflows/pull-tests-doc-indexer.yaml
+++ b/.github/workflows/pull-tests-doc-indexer.yaml
@@ -2,7 +2,7 @@ name: pull-tests-doc-indexer
 
 on:
   pull_request_target:
-    types: [ opened, synchronize, reopened, ready_for_review ]
+    types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - "main"
       - "release-**"
@@ -49,8 +49,6 @@ jobs:
           DOCS_PATH: "/test-data"
           DOCS_TABLE_NAME: "kc_pr_${{ github.event.pull_request.number }}"
         run: |
-          export CONFIG_PATH=$GITHUB_WORKSPACE/config/config.json
-          echo "${{ secrets.DOC_INDEXER_TESTS_CONFIG }}" | base64 --decode | jq > $CONFIG_PATH
           poetry run poe test-unit
 
       - name: Run integration tests
--- a/doc_indexer/src/main.py
+++ b/doc_indexer/src/main.py
@@ -55,6 +55,9 @@ def run_indexer() -> None:
     hana_conn = create_hana_connection(
         DATABASE_URL, DATABASE_PORT, DATABASE_USER, DATABASE_PASSWORD
     )
+    if not hana_conn:
+        logger.error("Failed to connect to the database. Exiting.")
+        raise
 
     indexer = AdaptiveSplitMarkdownIndexer(
         DOCS_PATH, embeddings_model, hana_conn, DOCS_TABLE_NAME
--- a/doc_indexer/src/utils/settings.py
+++ b/doc_indexer/src/utils/settings.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import sys
 from pathlib import Path
 
 from decouple import config
@@ -8,7 +9,6 @@ from decouple import config
 from utils.model_config import ModelConfig
 
 project_root = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../")
-
 MODELS_CONFIGS_KEY = "MODELS_CONFIGS"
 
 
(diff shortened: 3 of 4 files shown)


doc_indexer/src/utils/settings.py
```diff
@@ -49,23 +52,28 @@ def load_env_from_json() -> None:
 load_env_from_json()
 
 
-LOG_LEVEL = config("LOG_LEVEL", default="INFO")
-EMBEDDING_MODEL_NAME = config("EMBEDDING_MODEL_NAME", default="text-embedding-3-large")
+LOG_LEVEL = str(config("LOG_LEVEL", default="INFO"))
```
tanweersalah, line 55: why casting with str , when there is input parameter of cast available

doc_indexer/src/utils/settings.py
```diff
@@ -49,23 +52,28 @@ def load_env_from_json() -> None:
...
+DOCS_SOURCES_FILE_PATH = str(
+    config(
+        "DOCS_SOURCES_FILE_PATH",
+        default=os.path.join(project_root, "docs_sources.json"),
+    )
 )
-DOCS_PATH = config("DOCS_PATH", default="data")
-DOCS_TABLE_NAME = config("DOCS_TABLE_NAME", default="kyma_docs")
-CHUNKS_BATCH_SIZE = config("CHUNKS_BATCH_SIZE", cast=int, default=200)
+DOCS_PATH = str(config("DOCS_PATH", default="data"))
```
tanweersalah, line 67: why casting with str , when there is input parameter of cast available

doc_indexer/tests/integration/indexing/test_indexer.py
```diff
@@ -31,6 +31,12 @@ def hana_conn():
     hana_conn = create_hana_connection(
         DATABASE_URL, DATABASE_PORT, DATABASE_USER, DATABASE_PASSWORD
     )
+    if not hana_conn:
```
tanweersalah, line 34: same as above comment

doc_indexer/src/main.py
```diff
@@ -55,6 +55,9 @@ def run_indexer() -> None:
     hana_conn = create_hana_connection(
         DATABASE_URL, DATABASE_PORT, DATABASE_USER, DATABASE_PASSWORD
     )
+    if not hana_conn:
```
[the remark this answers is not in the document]
friedrichwilken: no, we do not properly handle exceptions for the creation of connections, we only log it:
```
    try:
        ...
        return connection
    except dbapi.Error:
        logger.exception("Connection to Hana Cloud failed.")
    except Exception:
        logger.exception("Unknown error occurred.")
    return None
```

and `create_hana_connecction` has the return type `dbapi.Connection | None` which is incompatible with AdaptiveSplitMarkdownIndexer which takes `dbapi.Connection`. This line makes ensures that hana_conn is only `dbapi.Connection` and not  `None`.

doc_indexer/src/utils/settings.py
friedrichwilken: Because `config` is, even with the `cast` argument, not able to set the type of the return value properly. If you set `cast` to `str` you will get `str | None` instead. But if we set defaults, the `None` part does not make sense, this can never return `None`. Casting it as `str` removes the `None` part.

doc_indexer/src/utils/settings.py
```diff
@@ -49,23 +52,28 @@ def load_env_from_json() -> None:
...
+DOCS_TABLE_NAME = str(config("DOCS_TABLE_NAME", default="kyma_docs"))
+CHUNKS_BATCH_SIZE = int(config("CHUNKS_BATCH_SIZE", cast=int, default=200))
 
-DATABASE_URL = config("DATABASE_URL")
-DATABASE_PORT = config("DATABASE_PORT", cast=int)
-DATABASE_USER = config("DATABASE_USER")
-DATABASE_PASSWORD = config("DATABASE_PASSWORD")
+DATABASE_URL = str(config("DATABASE_URL", default=""))
+DATABASE_PORT = int(config("DATABASE_PORT", cast=int, default=443))
```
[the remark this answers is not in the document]
friedrichwilken: Because `config` is, even with the `cast` argument, not able to set the type of the return value properly. If you set `cast` to `int` you will get `int | None` instead. But if we set defaults, the `None` part does not make sense, this can never return `None`. Casting it as `int` removes the `None` part.

doc_indexer/src/utils/settings.py
friedrichwilken: Because `config` is, even with the `cast` argument, not able to set the type of the return value properly. If you set `cast` to `str` you will get `str | None` instead. But if we set defaults, the `None` part does not make sense, this can never return `None`. Casting it as `str` removes the `None` part.

.github/workflows/pull-tests-doc-indexer.yaml
```diff
@@ -49,8 +49,6 @@ jobs:
           DOCS_PATH: "/test-data"
           DOCS_TABLE_NAME: "kc_pr_${{ github.event.pull_request.number }}"
         run: |
-          export CONFIG_PATH=$GITHUB_WORKSPACE/config/config.json
-          echo "${{ secrets.DOC_INDEXER_TESTS_CONFIG }}" | base64 --decode | jq > $CONFIG_PATH
```
tanweersalah, line 53: why removed ???

PR #903, opened 2025-11-25
refactor: use goal accuracy

friedrichwilken:
**Description**

 Refactored tests now use the same `Goal accuracy` metric that was introduced in PR #885, with evaluation criteria focused on:
  - Whether the core goal/intent is achieved
  - Allowing additional information that doesn't contradict the goal
  - Ensuring the main intent is clearly stated

 This should make the tests more robust and less brittle to variations in LLM output wording (stochasticity), while still ensuring the tests verify that the correct behavior is achieved.

**Related issue(s)**


elpadev (approved):
LGTM

Issue #930, opened 2025-12-08
Error handling in the Tools endpoint


danjiawork:
If we provide wrong credentials to the tool endpoint we get `Internal Server Error` , instead of proper response that the auth credential is invalid.

PR #939, opened 2025-12-11
fix: catch k8sClient errors properly by k8sClientError

danjiawork:
**Description**

This is an enhancement change for https://github.com/kyma-project/kyma-companion/pull/931

- Introduce a new class k8sClientError to include both context and error code, 
- wrap the tools’ error by the new class

**Related issue(s)**

https://github.com/kyma-project/kyma-companion/issues/930

### DIFF
--- a/src/agents/k8s/tools/query.py
+++ b/src/agents/k8s/tools/query.py
@@ -1,4 +1,3 @@
-from http import HTTPStatus
 from typing import Annotated
 
 from langchain_core.tools import tool
@@ -33,27 +32,17 @@ async def k8s_query_tool(
     The returned data is sanitized to remove any sensitive information.
     For example, it will always remove the `data` field of a `Secret` object."""
     try:
-        result = await k8s_client.execute_get_api_request(uri)
-        return result
+        return await k8s_client.execute_get_api_request(uri)
     except K8sClientError as e:
         # Add tool name if not already set
         if not e.tool_name:
             e.tool_name = "k8s_query_tool"
         raise
     except Exception as e:
-        # Extract status code if available (e.g., from ApiException)
-        # HTTPStatus members are ints, so we can safely use the value
-        status_code: int = (
-            e.status
-            if hasattr(e, "status") and isinstance(e.status, int)
-            else HTTPStatus.INTERNAL_SERVER_ERROR
-        )
-
-        raise K8sClientError(
-            message=str(e),
-            status_code=status_code,
-            uri=uri,
+        raise K8sClientError.from_exception(
+            exception=e,
             tool_name="k8s_query_tool",
+            uri=uri,
         ) from e
 
 
--- a/src/agents/kyma/tools/query.py
+++ b/src/agents/kyma/tools/query.py
@@ -1,4 +1,3 @@
-from http import HTTPStatus
 from typing import Annotated
 
 from langchain_core.tools import tool
@@ -35,27 +34,17 @@ async def kyma_query_tool(
     - /apis/serverless.kyma-project.io/v1alpha2/namespaces/default/functions
     - /apis/gateway.kyma-project.io/v1beta1/namespaces/default/apirules"""
     try:
-        result = await k8s_client.execute_get_api_request(uri)
-        return result
+        return await k8s_client.execute_get_api_request(uri)
     except K8sClientError as e:
         # Add tool name if not already set
         if not e.tool_name:
             e.tool_name = "kyma_query_tool"
         raise
     except Exception as e:
-        # Extract status code if available (e.g., from ApiException)
-        # HTTPStatus members are ints, so we can safely use the value
-        status_code: int = (
-            e.status
-            if hasattr(e, "status") and isinstance(e.status, int)
-            else HTTPStatus.INTERNAL_SERVER_ERROR
-        )
-
-        raise K8sClientError(
-            message=str(e),
-            status_code=status_code,
-            uri=uri,
+        raise K8sClientError.from_exception(
+            exception=e,
             tool_name="kyma_query_tool",
+            uri=uri,
         ) from e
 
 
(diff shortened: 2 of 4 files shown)


src/agents/k8s/tools/query.py
```diff
@@ -30,13 +31,35 @@ async def k8s_query_tool(
...
+        result = await k8s_client.execute_get_api_request(uri)
+        if not isinstance(result, (list, dict)):
+            raise K8sClientError(
+                message=f"Invalid result type: {type(result)}",
+                status_code=500,
+                uri=uri,
+                tool_name="k8s_query_tool",
+            )
+        return result
+    except K8sClientError as e:
```
ruanxin: this `K8sClientError` is defined by us, we know where it gets raised in the codebase, so it's better to assign the `tool_name` in each place when this error gets raised, instead of mutating the error in the exception logic and raising it again.

src/agents/kyma/tools/query.py
```diff
@@ -76,7 +95,10 @@ def fetch_kyma_resource_version(
...
         return resource_version
     except Exception as e:
-        raise Exception(
-            f"failed executing fetch_kyma_resource_version with resource_kind: {resource_kind},"
-            f" raised the following error: {e}"
+        # Convert exceptions to K8sClientError, preserving status code if available
+        raise K8sClientError(
+            message=str(e),
+            status_code=getattr(e, "status", 500),
+            uri="",
```
ruanxin: if there is no `uri`, you don't need to declare it.

src/agents/kyma/tools/query.py
```diff
@@ -76,7 +95,10 @@ def fetch_kyma_resource_version(
         resource_version = k8s_client.get_resource_version(resource_kind)
         return resource_version
     except Exception as e:
-        raise Exception(
-            f"failed executing fetch_kyma_resource_version with resource_kind: {resource_kind},"
-            f" raised the following error: {e}"
+        # Convert exceptions to K8sClientError, preserving status code if available
+        raise K8sClientError(
+            message=str(e),
+            status_code=getattr(e, "status", 500),
```
ruanxin: same as here: https://github.com/kyma-project/kyma-companion/pull/939/files#r2613929111

src/agents/k8s/tools/query.py
```diff
@@ -30,13 +31,35 @@ async def k8s_query_tool(
...
-            f"failed executing k8s_query_tool with URI: {uri}."
-            f"The result is not a list or dict, but a {type(result)}"
-        )
-    return result
+    try:
+        result = await k8s_client.execute_get_api_request(uri)
+        if not isinstance(result, (list, dict)):
+            raise K8sClientError(
+                message=f"Invalid result type: {type(result)}",
+                status_code=500,
```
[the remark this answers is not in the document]
danjiawork: Agree, will replace all 500 with `HTTPStatus.INTERNAL_SERVER_ERROR` throughout the codebase in the next commit

src/agents/k8s/tools/query.py
danjiawork: The K8sClientError with empty tool_name from the client layer is intentional, it allows the client to be reusable across different tools, and each tool explicitly sets its own tool_name when handling the error. This provides better separation of concerns.

src/services/k8s.py
```diff
@@ -27,6 +27,47 @@
 GROUP_VERSION_PARTS_COUNT = 2
 
 
+class K8sClientError(Exception):
```
ruanxin: I don't think it's a good idea put this Error class under `services`, when you import the exception, you will break dependency flow if designed like this. For example, ideally we don't need to expose `services` layer to `routers` at all (even though the current structure already violates this, but we can refactor later), The router should only need to interact with the tools layer.

However, by placing `K8sClientError` inside services, we’re forcing the routers layer to import it and then cast it to HTTPException, which creates an unnecessary hard dependency from routers to services.

My suggestion is to move the custom exception class into `utils` package instead. A utility module is typically meant for cross-layer, shared objects, which fits this use case much better.

src/agents/k8s/tools/query.py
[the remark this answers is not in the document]
danjiawork: addressed in [6133f92](https://github.com/kyma-project/kyma-companion/pull/939/commits/6133f92395cb7afae70c4533fdf61caf5f24cdc0)

src/agents/k8s/tools/query.py
```diff
@@ -30,13 +31,35 @@ async def k8s_query_tool(
...
+        return result
+    except K8sClientError as e:
+        if not e.tool_name:
+            e.tool_name = "k8s_query_tool"
+        raise
+    except ApiException as e:
+        # Preserve ApiException status code in K8sClientError
+        raise K8sClientError(
+            message=str(e),
+            status_code=e.status,
```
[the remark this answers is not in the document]
danjiawork: addressed in [6133f92](https://github.com/kyma-project/kyma-companion/pull/939/commits/6133f92395cb7afae70c4533fdf61caf5f24cdc0)

src/agents/kyma/tools/query.py
```diff
@@ -34,16 +35,34 @@ async def kyma_query_tool(
...
+            uri=uri,
+            tool_name="kyma_query_tool",
+        ) from e
     except Exception as e:
-        raise Exception(
-            f"failed executing kyma_query_tool with URI: {uri},raised the following error: {e}"
+        # Convert other exceptions to K8sClientError
+        raise K8sClientError(
+            message=str(e),
+            status_code=getattr(e, "status_code", 500),
```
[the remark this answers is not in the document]
danjiawork: addressed in [6133f92](https://github.com/kyma-project/kyma-companion/pull/939/commits/6133f92395cb7afae70c4533fdf61caf5f24cdc0)

src/agents/kyma/tools/query.py
danjiawork: addressed in [6133f92](https://github.com/kyma-project/kyma-companion/pull/939/commits/6133f92395cb7afae70c4533fdf61caf5f24cdc0)

src/agents/k8s/tools/query.py
```diff
@@ -30,13 +32,29 @@ async def k8s_query_tool(
...
+        return result
+    except K8sClientError as e:
+        # Add tool name if not already set
+        if not e.tool_name:
+            e.tool_name = "k8s_query_tool"
+        raise
+    except Exception as e:
+        # Extract status code if available (e.g., from ApiException)
+        # HTTPStatus members are ints, so we can safely use the value
+        status_code: int = (
```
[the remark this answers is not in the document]
ruanxin: btw, what if `e` doesn't have attribute `status`? then the line 47 `e.status` will raise expection or?

ruanxin (approved):

src/utils/exceptions.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        # Extract status code if available (e.g., from ApiException)
+        # ApiException.status can be either an int or a string
+        status_code: int = HTTPStatus.INTERNAL_SERVER_ERROR
+        if hasattr(exception, "status"):
+            status = exception.status
+            if isinstance(status, int) or (
+                isinstance(status, str) and status.isdigit()
+            ):
+                status_code = int(status)
```
mfaizanse: validate if it is a valid HTTP status code.

danjiawork: Enhanced in https://github.com/kyma-project/kyma-companion/pull/945

PR #991, opened 2026-01-21
refactor: replace black, flake8 and flake8-bugbear with ruff

friedrichwilken:
**Description**

Our dev dependencies currently include both Ruff (for linting) and Black (for formatting), plus Flake8 and flake8-bugbear for additional linting checks. This creates redundancy since Ruff is a drop-in replacement for all three tools. This PR consolidates our code quality tooling to use only Ruff for both linting and formatting.

  Benefits:
  - Faster execution (Ruff is [10-100x faster than Black/Flake8](https://github.com/astral-sh/ruff/blob/main/README.md))
  - Simplified tooling with single tool for linting and formatting
  - Maintains the olds, Black-compatible, formatting style
  - Reduces dev dependencies by 3 packages
  - Reduces maintenance burden with fewer dependencies to update
  - All existing poe tasks continue to work identically

The same changes have been applied to the doc_indexer and the blackbox subprojects.
  
Note: we configured ruff to format with a line length of 120 characters. However, we never actually used ruff for formatting. Instead, we used black to format, which we never configured and so it was formatting to its default line length of 88 characters. With the changes to the formatters in this PR, there will be a lot of re-formatting to our inteded line length of 120 characters.
 
**Related issue(s)**


tanweersalah (approved):
lgtm

PR #1005, opened 2026-02-04
refactor: code quality

friedrichwilken:
**Description**

- reduce imports: instead of importing `langchain[all]` and `langgraph[all]`, let's only import packages we use
- remove httpx deprecation warning by using the `content` field instead of `data`
- clean up the `pyproject.toml`: at the moment we still have a mix of `poetry v1` and `poetry v2` in our `pyproject.toml`
- fix name of file  `contants.py` to `constants.py`
- bump docker base image

**Related issue(s)**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/gardenlinux/gardenlinux:1877.10 AS builder
+FROM ghcr.io/gardenlinux/gardenlinux:1877.11 AS builder
 WORKDIR /app
 
 # Copy only necessary files


doc_indexer/pyproject.toml
```diff
@@ -24,7 +28,6 @@ dependencies = [
 ]
 
 [tool.poetry]
-packages = [{ include = "src" }]
```
mfaizanse: why remove?

pyproject.toml
```diff
@@ -33,7 +38,6 @@ dependencies = [
   "uvicorn (>=0.40.0,<0.41.0)"
 ]
 [tool.poetry]
-packages = [{ include = "src" }]
```
mfaizanse: why remove?

PR #1020, opened 2026-02-19
fix: pod logs tool return 404 if nothing was found

friedrichwilken:
**Description**

- if the pod logs tool cannot find anything, we will return 404

### DIFF
--- a/src/utils/exceptions.py
+++ b/src/utils/exceptions.py
@@ -1,9 +1,5 @@
 import json
 from http import HTTPStatus
-from typing import TYPE_CHECKING
-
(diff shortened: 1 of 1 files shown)


src/utils/exceptions.py
```diff
@@ -19,6 +23,29 @@ def parse_k8s_error_response(error_text: str) -> str:
         return error_text
 
 
+class InvalidContainerError(Exception):
```
danjiawork: looks like this is defined but never used.

friedrichwilken: that was removed earlier.

PR #1061, opened 2026-03-10
fix: handle metrics server availability and fix type casting bugs

friedrichwilken:
**Description**

- ...
- ...
- ...

**Related issue(s)**


friedrichwilken:
✅ Addressed review comment from @hyperspace-insights

Added warning log when the data sanitizer returns an unexpected type for events (commit d3dd767). This prevents silent data loss and makes debugging easier.

The change:
- Previously: silently returned `[]` when sanitizer returned non-list
- Now: logs a warning with the actual type before returning `[]`

This helps operators identify configuration or data sanitization issues quickly.

PR #1105, opened 2026-04-08
fix: doc indexer workflow fails if PR author is not a codeowner

friedrichwilken:
## Description

- `pull-tests-doc-indexer.yaml`: workflow now fails explicitly for non-codeowners instead of silently skipping. Added `ok-to-test` label as a codeowner override - if the label is present, the membership check is bypassed.
- `pull-request-comment.yaml`: checks team membership on PR open and posts a tailored comment - codeowners get the standard notes, external contributors get an explanation of the authorization model and how to request the `ok-to-test` label.
- `github-comment.yaml`: split `pr-notes` into `pr-notes-codeowner` and `pr-notes-external` templates.

### DIFF
--- a/.github/github-comment.yaml
+++ b/.github/github-comment.yaml
@@ -1,5 +1,5 @@
 post:
-  pr-notes:
+  pr-notes-codeowner:
     template: |
       ## Note(s) for PR Author:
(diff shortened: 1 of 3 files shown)


danjiawork (commented):
Looks good to me overall, only left one small suggestion

danjiawork (approved):

PR #1138, opened 2026-04-20
test: verify indexer workflow path filters

friedrichwilken:
Test PR to verify that the indexer build and e2e workflows trigger correctly on `doc_indexer/**` changes after #1131 merged.

Close after checks pass (or fail — result goes into the task doc).


friedrichwilken:
Closing — this was a test PR to verify the indexer path filters from #1131. The  failures were caused by the pagination bug fixed in #1139 (now merged). Results confirmed path filters work correctly.

PR #1139, opened 2026-04-20
fix: paginate check-run API calls in wait-for-commit-check script

friedrichwilken:
## Problem

`scripts/python/wait-for-commit-check/run.py` fetched only the first page (default 30 items) of check runs from the GitHub API. On PRs with many parallel CI jobs (kyma-companion has 52 check runs per commit), the target check run could appear beyond position 30 and would be missed — causing the `Wait for image build job` step to time out after 900 s.

The API does not guarantee ordering, so this is an intermittent failure depending on whether the target check run happens to land on page 1.

This was discovered via PR #1138 (test PR for #1131): `build-indexer / Build image` completed successfully but the poller timed out because it was on page 2.

**Proof:**
```
gh api "repos/kyma-project/kyma-companion/commits/052364abf4c9db1bae2d681390b7fbfb1b14276d/check-runs" --jq '"page1: \(.check_runs|length) / total: \(.total_count)"'
# page1: 30 / total: 52
```

## Fix

- Add `per_page=100` to the initial API request (GitHub max).
- Follow `Link: rel="next"` pagination headers to accumulate all check runs across pages.

Affects all four workflows that use the script:
- `pull-e2e-tests-doc-indexer.yaml`
- `push-e2e-tests-doc-indexer.yaml`
- `pull-api-tests.yaml`
- `pull-evaluation-tests.yaml`

## Testing

After this fix merges, PR #1138 will be re-run to confirm.

Related: #1130, #1131, #1138

PR #1140, opened 2026-04-20
refactor: extract wait-for-build into a composite action

friedrichwilken:
## What

Introduces `.github/actions/wait-for-build/action.yml` — a composite action that replaces the four-step wait-for-build pattern with a single bash step using `gh api` and `jq`. Migrates `push-e2e-tests-doc-indexer.yaml` to use it as a first step.

The remaining three pull workflows are migrated in a follow-up PR after this one is confirmed working end-to-end.

## Why

The four-step pattern (checkout → setup-python → pip install → run script) was copy-pasted across all four workflows. The bash rewrite drops all three setup steps — `gh` and `jq` are pre-installed on GitHub-hosted runners — and `gh api --paginate` handles pagination natively, so the per-page/Link-header logic from #1139 comes for free.

## Testing

This is a composite action change referenced as `@main`, so it cannot be exercised until this PR merges. A test PR will be opened immediately after merge to verify the action runs correctly end-to-end before the follow-up PR is created.

### DIFF
--- a/.github/actions/wait-for-build/action.yml
+++ b/.github/actions/wait-for-build/action.yml
@@ -1,0 +1,49 @@
+name: "Wait for build"
+description: "Waits for a GitHub check run to complete successfully before proceeding."
+
+inputs:
+  git-ref:
+    description: "The commit SHA to poll."
+    required: true
+  check-run-name:
+    description: "The name of the check run to wait for (e.g. 'build / Build image')."
+    required: true
+  interval:
+    description: "Polling interval in seconds."
+    required: false
+    default: "60"
+  timeout:
+    description: "Maximum wait time in seconds."
+    required: false
+    default: "900"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout code
+      uses: actions/checkout@v6
+      with:
+        ref: main
+
+    - name: Setup python
+      uses: actions/setup-python@v6
+      with:
+        python-version: "3.10"
+        cache: "pip"
+
+    - name: Install requirements
+      shell: bash
+      run: |
+        pip install -r ./scripts/python/wait-for-commit-check/requirements.txt
+
+    - name: Wait for build
+      shell: bash
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+        GIT_REF: ${{ inputs.git-ref }}
+        REPOSITORY_FULL_NAME: ${{ github.repository }}
+        GIT_CHECK_RUN_NAME: ${{ inputs.check-run-name }}
+        INTERVAL: ${{ inputs.interval }}
+        TIMEOUT: ${{ inputs.timeout }}
+      run: |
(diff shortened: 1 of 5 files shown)


.github/actions/wait-for-build/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    required: false
+    default: "900"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout code
+      uses: actions/checkout@v6
+      with:
+        ref: main
```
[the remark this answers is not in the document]
friedrichwilken: ref: main is intentional. the script lives in the repo and we always want the main version - same as the original inline steps did. actually its a feature: once #1139 lands on main, all callers pick up the pagination fix automatically without any changes. binding it to github.action_ref would add no benefit here since the action is always called @main anyway

.github/actions/wait-for-build/action.yml
```diff
@@ -0,0 +1,83 @@
...
+        GIT_CHECK_RUN_NAME: "${{ inputs.check-run-name }}"
+        INTERVAL: "${{ inputs.interval }}"
+        TIMEOUT: "${{ inputs.timeout }}"
+      run: |
+        echo "Repository:     $REPOSITORY_FULL_NAME"
+        echo "Git ref:        $GIT_REF"
+        echo "Check run name: $GIT_CHECK_RUN_NAME"
+        echo "Timeout:        ${TIMEOUT}s / Interval: ${INTERVAL}s"
+
+        start=$(date +%s)
```
mfaizanse, line 38: Why not reuse the python [script](https://github.com/kyma-project/kyma-companion/tree/main/scripts/python/wait-for-commit-check) which we already have? Not a big fan of merging code in workflow files.

.github/actions/wait-for-build/action.yml
```diff
@@ -0,0 +1,83 @@
...
+    description: "The name of the check run to wait for (e.g. 'build / Build image')."
+    required: true
+  interval:
+    description: "Polling interval in seconds."
+    required: false
+    default: "60"
+  timeout:
+    description: "Maximum wait time in seconds."
+    required: false
+    default: "900"
```
mfaizanse, line 18: please mentioned list of required secrets here as well. Lets not pass all secrets from the calling workflow, rather only pass secrets which are required here.

.github/actions/wait-for-build/action.yml
friedrichwilken: That was my first approach, but I guess there are two topics here:
1. Vice versa, I prefer the idea of a "self contained" workflow, that does not rely on external dependencies like scripts, especially when these are not shared between multiple workflows anyhow. I guess for python that would be the right way, but 
2. I think python is just not the optimal tool here. Since this is all about the github api, this task can be done easiely with the gh cli tool with far less complexity.

PR #1144, opened 2026-04-23
feat: accept raw key bytes in KeyStore, align key generation to cert-manager PEM format

### LINKED ISSUE #1143: support PEM private key files
## Background

The companion encryption service previously loaded its EC private key only from a base64-encoded environment variable (`ENCRYPTION_PRIVATE_KEY_B64`). cert-manager, which provisions TLS certificates in Kubernetes, stores private keys as PEM-encoded PKCS8 files in a mounted secret (`tls.key`). There was no way to load the key directly from such a file.

## What was needed

1. **Support PEM (and DER) private key files** — The key loading logic needed to accept raw bytes read from a file instead of a base64-encoded string, and transparently handle both PEM and DER encodings.

2. **A `KeyStore` singleton** — Centralise key management in a single place that:
   - Loads the EC private key from the file path configured via `ENCRYPTION_PRIVATE_KEY_PATH` (the Kubernetes-mounted cert-manager secret).
   - Falls back to the existing `ENCRYPTION_PRIVATE_KEY_B64` env var when no path is set.
   - Automatically reloads the key from disk every 30 minutes so cert-manager key rotations are picked up without a pod restart.
   - Exposes `get_private_key()`, `get_public_key()`, and `get_public_key_str()` helpers.

3. **Decouple key decoding from `Encryption`** — `Encryption.__init__` should receive a ready `EllipticCurvePrivateKey` object rather than a raw base64 string, keeping concerns separated.

4. **Align the key-generation script** — `generate-ecdh-keys.sh` should produce a single `tls.key` file in PEM/PKCS8 format, identical to what cert-manager writes, for local development and testing.

5. **Update routers** — `routers/common.py` and `routers/public_key.py` should obtain the key via `KeyStore()` instead of reading the raw env var directly.

mfaizanse:
Closes #1143

## Summary

- `_load_private_key` now accepts `bytes` (PEM or DER) directly instead of a base64-encoded string; callers handle the base64 decode themselves
- `_load_private_key_from_file` uses `read_bytes()` so binary DER files are supported alongside PEM files
- `generate-ecdh-keys.sh` simplified to produce a single `tls.key` (PEM/PKCS8), matching the format that cert-manager writes into Kubernetes secrets
- `Encryption.__init__` now receives an `EllipticCurvePrivateKey` directly (key decoding moved to `KeyStore`); tests and mocks updated accordingly
- `ENCRYPTION_PRIVATE_KEY_PATH` / `ENCRYPTION_PRIVATE_KEY_B64` settings now have explicit `cast=str`
- Adds `tests/unit/services/test_key_store.py` covering `_load_private_key`, `_load_private_key_from_file`, singleton behaviour, key reload interval, and the B64 config fallback path

## Test plan

- [ ] `pytest tests/unit/services/test_key_store.py` — new key store unit tests pass
- [ ] `pytest tests/unit/services/test_encryption.py` — updated encryption tests pass
- [ ] `pytest tests/unit/routers/test_common.py tests/unit/routers/test_public_key.py` — router tests pass
- [ ] Run `./scripts/shell/generate-ecdh-keys.sh` and verify `tls.key` is produced as a valid PEM/PKCS8 file

### DIFF
--- a/src/services/key_store.py
+++ b/src/services/key_store.py
@@ -1,0 +1,180 @@
+"""
+Singleton key store that loads an EC private key from a Kubernetes-mounted
+secret file and transparently reloads it when the cached copy is stale.
+"""
+
+import base64
+import time
+from pathlib import Path
+
+from cryptography.hazmat.primitives.asymmetric.ec import (
+    EllipticCurvePrivateKey,
+    EllipticCurvePublicKey,
+)
+from cryptography.hazmat.primitives.serialization import (
+    Encoding,
+    PublicFormat,
+    load_der_private_key,
+    load_pem_private_key,
+)
+
+from utils.logging import get_logger
+from utils.settings import ENCRYPTION_PRIVATE_KEY_B64, ENCRYPTION_PRIVATE_KEY_PATH
+from utils.singleton_meta import SingletonMeta
+
+logger = get_logger(__name__)
+
+_RELOAD_INTERVAL_SECONDS = 30 * 60  # 30 minutes
+
+
+def _load_private_key(data: bytes) -> EllipticCurvePrivateKey:
+    """Decode key data and return a validated EC private key.
+
+    Supports both PEM and raw DER formats (after base64 decoding).
+
+    :param data: Raw private key bytes in PEM or DER format.
+    :raises TypeError: If the key is not an EC private key.
+    :raises ValueError: If the data cannot be decoded or parsed as a private key.
+    """
+
+    # Try PEM first (has header line), fall back to DER.
+    try:
+        key = load_pem_private_key(data, password=None)
+    except (ValueError, TypeError):
+        key = load_der_private_key(data, password=None)
+
+    if not isinstance(key, EllipticCurvePrivateKey):
+        raise TypeError("Key is not an EC private key")
+    return key
+
+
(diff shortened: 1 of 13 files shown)


src/services/key_store.py
```diff
@@ -0,0 +1,180 @@
...
+    :param data: Raw private key bytes in PEM or DER format.
+    :raises TypeError: If the key is not an EC private key.
+    :raises ValueError: If the data cannot be decoded or parsed as a private key.
+    """
+
+    # Try PEM first (has header line), fall back to DER.
+    try:
+        key = load_pem_private_key(data, password=None)
+    except (ValueError, TypeError):
+        key = load_der_private_key(data, password=None)
```
friedrichwilken, line 45: what happens here, if the PEM file is malformatted (not sure if that is a realistic scenario, given that it would have been created by the cert-manager)?
we would catch that err and instead would try to read a DER file, which --in case of an actual malformatted PEM file-- would not exist. now `load_der_private_key` would return an err that we would not catch.

so in that case we would return some err about the DER file, not about the PEM file. 
I guess we would know that the problem is the PEM and that we do not have a DER file, but it would still be misleading, I think.

mfaizanse: added a log line to print the exception for PEM file.

friedrichwilken (approved):

PR #1198, opened 2026-06-02
fix: avoid exposing exception details in logs and HTTP responses

ruanxin:
## Summary

- Replace f-string exception interpolation (`f"... {e}"`) with `exc_info=e` in logger calls so the logging framework controls exception output rather than embedding raw exception strings in messages
- Remove `str(e)` from `HTTPException` detail fields to prevent internal error details from leaking to API clients
- Affects `src/services/hana.py` and `src/routers/kyma_tools_api.py`

Addresses Checkmarx SAST findings (severity: LOW):
- `Information_Exposure_Through_an_Error_Message` in `src/services/hana.py` (lines 37, 40, 76)
- `Information_Exposure_Through_an_Error_Message` in `src/routers/kyma_tools_api.py` (lines 75, 111, 117)

## Test plan

- [ ] Verify existing unit tests still pass
- [ ] Confirm logger still captures full exception tracebacks via `exc_info`
- [ ] Confirm HTTP error responses return generic messages without internal details


src/services/hana.py
```diff
@@ -34,10 +34,10 @@ def __init__(self, connection_factory: Callable[[], dbapi.Connection] | None = N
         try:
             self.connection = connection_factory() if connection_factory else _get_hana_connection()
         except dbapi.Error as e:
-            logger.error(f"Connection to Hana Cloud failed: {e}")
+            logger.error("Connection to Hana Cloud failed.", exc_info=e)
```
tanweersalah: can we use `logger.exception` instead of `logger.error` where we need the traces

src/services/hana.py
```diff
@@ -34,10 +34,10 @@ def __init__(self, connection_factory: Callable[[], dbapi.Connection] | None = N
...
             self.connection = None
         except Exception as e:
-            logger.error(f"Unknown error occurred: {e}")
+            logger.error("Unknown error occurred.", exc_info=e)
```
tanweersalah: can we use `logger.exception` instead of `logger.error` where we need the traces

src/services/hana.py
```diff
@@ -73,7 +73,7 @@ def is_connection_operational(self) -> bool:
             self._last_health_state = True
             return True
         except Exception as e:
-            logger.error(f"Error while connecting to HANA DB: {e}")
+            logger.error("Error while connecting to HANA DB.", exc_info=e)
```
tanweersalah: can we use `logger.exception` instead of `logger.error` where we need the traces

tanweersalah (approved):

PR #1206, opened 2026-06-10
chore: upgrade ECDH curve to P-521 and simplify key loading

mfaizanse:
## Summary

- Upgrades the ECDH curve from P-256 (`secp256r1`) to P-521 (`secp521r1`) for stronger key exchange security; updates HKDF hash from SHA-256 to SHA-384 to match.
- Removes the `ENCRYPTION_PRIVATE_KEY_B64` fallback — the private key must now be loaded from a PEM file at `ENCRYPTION_PRIVATE_KEY_PATH` (defaults to `config/encryption_key.pem`). (This was done for backward compatibility for the already released version, but now the latest released version do not need this)
  - Removes raw DER key format support; only PEM is accepted now.
  - Removes raw private key files from the repo and adds them to `.gitignore`.
- Updates the key-generation script to produce a P-521 key written to `config/encryption_key.pem`.

### DIFF
--- a/scripts/shell/generate-ecdh-keys.sh
+++ b/scripts/shell/generate-ecdh-keys.sh
@@ -1,31 +1,29 @@
 #!/usr/bin/env bash
-# Generate an ECDSA P-256 private key matching the cert-manager Certificate CR
-# (ECDSA, P-256, PKCS8 encoding).
+# Generate an ECDSA P-521 private key for ECDH key exchange
+# (ECDSA, P-521/secp521r1, PKCS8 encoding).
 #
-# cert-manager stores the key as PEM (PKCS8) in the tls.key field of the
-# Kubernetes secret.  This script produces the same format for local
+# This script produces a PEM (PKCS8) encoded EC private key for local
 # development / testing.
 #
 # Output file:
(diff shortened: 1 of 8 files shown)


scripts/shell/generate-ecdh-keys.sh
```diff
@@ -1,31 +1,29 @@
...
 #
 # Usage:
 #   ./generate-ecdh-keys.sh [output_dir]
-#   output_dir defaults to ./ecdh-keys
+#   output_dir defaults to ./config
 
 set -euo pipefail
 
-OUTPUT_DIR="${1:-./ecdh-keys}"
-mkdir -p "$OUTPUT_DIR"
```
friedrichwilken, line 19: Why did you remove this one? wouldnt it fail if the dir does not exist?

mfaizanse: I changed the default to "./config" DIR. "./config" DIR will exists because we have the config.json in this directory anyway.

friedrichwilken: fair!

friedrichwilken (approved):

PR #1213, opened 2026-06-11
test: add TLS-enabled Redis to evaluation test pipeline

friedrichwilken:
## Summary

- Replace the `redis-stack` Helm chart (RSAL-licensed, no TLS support) with a plain
  `redis:alpine` deployment configured with TLS
- Add `scripts/k8s/create-redis-tls-certs.sh` -- generates a self-signed CA + server
  cert; can be run locally for testing
- Add `scripts/k8s/redis-tls-manifest.yaml` -- Namespace, ConfigMap, Deployment, Service
  using `redis:alpine` with TLS on port 6380
- Mount the Redis CA cert into the companion pod and set `REDIS_SSL_ENABLED=true` in the
  k3d manifest -- exercises the TLS connection path in CI

This is a prerequisite for enforcing TLS 1.3 on the Redis connection. Once this merges,
the enforcement PR will be validated end-to-end in CI before it merges.

## Testing

- `create-redis-tls-certs.sh` tested locally -- generates valid certs accepted by Python 3.13
- Redis TLS deployment tested on k3d (`kyma-companion-test`) -- TLS 1.3 and TLS 1.2 both
  connect successfully from inside the cluster

### DIFF
--- a/scripts/k8s/create-redis-tls-certs.sh
+++ b/scripts/k8s/create-redis-tls-certs.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -euo pipefail
+
+OUTPUT_DIR="${1:-/tmp/redis-tls}"
+mkdir -p "$OUTPUT_DIR"
+
+# CA key + cert with proper key usage (required by Python 3.13 strict X.509 validation)
+openssl genrsa -out "$OUTPUT_DIR/ca.key" 4096
+openssl req -new -x509 -days 365 \
+  -key "$OUTPUT_DIR/ca.key" \
+  -out "$OUTPUT_DIR/ca.crt" \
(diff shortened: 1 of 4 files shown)


scripts/k8s/create-redis-tls-certs.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
```
mfaizanse, line 2: Can you add comments about the purpose of this script?

mfaizanse (commented):
Whats the goal of this test?
To check if it is using TLS1.3 to connect to redis?
Or to just check if it works with cert and ssl?
If someone changes the TLS1.3 config back to default, will the test fail?

PR #1224, opened 2026-06-18
feat: added new endpoint for fetching cluster region

tanweersalah:
# Add Cluster Region Endpoint for SKR Region Lookup

### New Feature

✨ Introduces a new REST API endpoint `GET /api/tools/cluster-region/{shoot_id}` that returns cloud provider region information for a given SKR (Shoot Kubernetes Runtime) cluster. The service reads Runtime Custom Resources from the KCP cluster using the pod's in-cluster service account, with results cached in Redis for 3 days.

### Changes

* `src/routers/cluster_region_api.py` *(new)*: Defines the FastAPI router for the new endpoint. Delegates to the `get_cluster_region` service and handles HTTP exceptions with structured error responses.

* `src/services/cluster_region.py` *(new)*: Core service logic that:
  - Checks Redis cache for a previously fetched result (TTL: 3 days)
  - On cache miss, queries the KCP cluster's Runtime CR using a label selector on `kyma-project.io/shoot-name`
  - Extracts `region`, `platformRegion`, and `provider` labels and computes the `isEUAccessOnly` flag based on known EU-Access platform regions (`cf-eu11`, `cf-ch20`, `cf-eu01`, `cf-eu02`, `cf-eu31`)
  - Caches the result and returns a `ClusterRegionResponse`

* `src/routers/common.py`: Added the `ClusterRegionResponse` Pydantic model with aliased fields (`shoot-id`, `platformRegion`, `isEUAccessOnly`) and `populate_by_name` enabled.

* `src/main.py`: Registered the new `cluster_region_router` with the FastAPI app.

* `tests/unit/routers/test_cluster_region_api.py` *(new)*: Unit tests covering successful responses, EU-access flag logic, 404 handling, unexpected error handling, and cache hit/miss behavior.

* `tests/blackbox/src/api_tests/test_cluster_region_api.py` *(new)*: API/integration tests validating the endpoint against Runtime CRs installed in a k3d test cluster, covering general runtimes, EU-access runtimes, unknown shoot IDs (404), and response schema validation.

- [ ] 🔄 Regenerate and Update Summary

<details>
<summary>PR Bot Information</summary>

**Version:** `1.26.5`

- Correlation ID: `ea5133d9-0ba9-4f5f-9d06-94a4b98137b4`
- LLM: `anthropic--claude-4.6-sonnet`
- Summary Prompt: [Default Prompt](https://github.tools.sap/intelligent-insights/i2-pull-request/blob/main/src/services/llm/prompts/summary_instructions_prompt.md)
- Output Template: [Default Template](https://github.tools.sap/intelligent-insights/i2-pull-request/blob/main/src/services/llm/prompts/summary_default_output_template.md)
- File Content Strategy: Full file content
- Event Trigger: `pull_request.edited`
</details>

### DIFF
--- a/src/services/cluster_region.py
+++ b/src/services/cluster_region.py
@@ -6,7 +6,6 @@ Fetches the Runtime CR for a given shoot-id from the local Kubernetes cluster
 in Redis for 3 days.
 """
 
-import json
 from http import HTTPStatus
 
 from kubernetes import client, config, dynamic
@@ -89,7 +88,7 @@ async def get_cluster_region(shoot_id: str) -> ClusterRegionResponse:
             cached = await redis.get_connection().get(_cache_key(shoot_id))
             if cached:
                 logger.debug(f"Cache hit for shoot_id={shoot_id}")
-                return ClusterRegionResponse(**json.loads(cached))
+                return ClusterRegionResponse.model_validate_json(cached)
         except Exception:
             logger.warning(f"Redis read failed for shoot_id={shoot_id}, proceeding without cache")
 
--- a/src/services/cluster_region.py
+++ b/src/services/cluster_region.py
@@ -32,7 +32,6 @@ _KCP_NAMESPACE = "kcp-system"
 # Runtime CR group/version/plural
 _RUNTIME_GROUP = "infrastructuremanager.kyma-project.io"
 _RUNTIME_VERSION = "v1"
-_RUNTIME_PLURAL = "runtimes"
 
 # Redis cache TTL: 3 days in seconds
 _CACHE_TTL_SECONDS = 3 * 24 * 60 * 60
--- a/src/services/cluster_region.py
+++ b/src/services/cluster_region.py
@@ -60,11 +60,12 @@ def _build_response(shoot_id: str, labels: dict) -> ClusterRegionResponse:
 
 def _get_dynamic_client() -> dynamic.DynamicClient:
     """Load k8s config (in-cluster when deployed, local kubeconfig otherwise)."""
+    conf = client.Configuration()
     try:
-        config.load_incluster_config()
+        config.load_incluster_config(client_configuration=conf)
     except config.ConfigException:
-        config.load_kube_config()
-    return dynamic.DynamicClient(client.ApiClient())
+        config.load_kube_config(client_configuration=conf)
+    return dynamic.DynamicClient(client.ApiClient(configuration=conf))
 
 
 async def get_cluster_region(shoot_id: str) -> ClusterRegionResponse:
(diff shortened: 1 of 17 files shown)


src/services/cluster_region.py
```diff
@@ -0,0 +1,142 @@
...
+            "provider": labels.get(_LABEL_PROVIDER, ""),
+            "isEUAccessOnly": platform_region in _EU_ACCESS_PLATFORM_REGIONS,
+        }
+    )
+
+
+def _get_dynamic_client() -> dynamic.DynamicClient:
+    """Load k8s config (in-cluster when deployed, local kubeconfig otherwise)."""
+    try:
+        config.load_incluster_config()
```
mfaizanse: Can you add a variable in setting.py for KCP_KUBECONFIG and use it from there, so we can modify the path in different places like k3d or kcp.

tanweersalah: update to create seperate instance for very request

src/services/cluster_region.py
```diff
@@ -0,0 +1,143 @@
...
+# EU-Access platform regions
+_EU_ACCESS_PLATFORM_REGIONS = frozenset(["cf-eu11", "cf-ch20", "cf-eu01", "cf-eu02", "cf-eu31"])
+
+# KCP namespace where Runtime CRs live
+_KCP_NAMESPACE = "kcp-system"
+
+# Runtime CR group/version/plural
+_RUNTIME_GROUP = "infrastructuremanager.kyma-project.io"
+_RUNTIME_VERSION = "v1"
+_RUNTIME_PLURAL = "runtimes"
```
friedrichwilken: is this actually used?

src/services/cluster_region.py
```diff
@@ -0,0 +1,143 @@
...
+
+def _cache_key(shoot_id: str) -> str:
+    return f"{_CACHE_KEY_PREFIX}{shoot_id}"
+
+
+def _build_response(shoot_id: str, labels: dict) -> ClusterRegionResponse:
+    platform_region = labels.get(_LABEL_PLATFORM_REGION, "")
+    return ClusterRegionResponse(
+        **{
+            "shoot-id": shoot_id,
```
friedrichwilken, line 51: not sure if I get this wrong, but the original issue states:

> From the Runtime CR, create a response object:
> shoot-id: "<Value from label: kyma-project.io/shoot-name>"
so, will this work?

src/services/cluster_region.py
tanweersalah: removed

src/services/cluster_region.py
tanweersalah: yes , as discussed , both are same value , because we fetch the CRs label with this value only

friedrichwilken (approved):

PR #1289, opened 2026-07-21
chore: fix CodeQL quality findings

friedrichwilken:
## Description

Fixes the following CodeQL quality findings and underlying bugs:

**Statement has no effect (63 findings)** -- Removes redundant `...` placeholders from Protocol and ABC stub method bodies across `src/`. Methods with a docstring need no additional placeholder body. `PIE790` (unnecessary placeholder) is already enforced via the `PIE` group in `ruff.toml`; a clarifying comment is added inline.

**Wrong number of arguments in a class instantiation (36 findings)** -- `K8sClient()` was called without the required `k8s_auth_headers` argument in tests that patch `__init__` to `return None`. Fixed by passing `Mock()` as a placeholder argument.

**Use of exit() or quit() (11 findings)** -- Replaces `exit()` with `sys.exit()` in `scripts/python/wait-for-commit-check/run.py`. `exit()` depends on the `site` module being loaded; `sys.exit()` is always available.

**Membership test with a non-container (10 findings)** -- Test lambdas used `"Error" in msg` where CodeQL could not prove `msg` is a container. Fixed by using `"Error" in str(msg)`.

**Implicit string concatenation in a list (3 findings)** -- Two evaluation steps in integration tests (`test_k8s_agent.py`, `test_kyma_agent.py`) were silently concatenated into one string due to a missing comma, reducing the number of evaluated criteria from 2 to 1. Fixed by adding the missing comma. The third was an intentional multi-line string literal, rewritten with explicit `+` concatenation.

**Unused global variable (3 findings)** -- `_logging_configured` in `src/utils/logging.py` and `doc_indexer/src/utils/logging.py` was flagged because it was only accessed via the `global` keyword inside functions. Replaced with a `list[bool]` so mutation does not require `global`. Also removes dead `config_path` variable in `scripts/python/k8s_resources_discovery.py`.

**Module is imported with 'import' and 'import from' (2 findings)** -- Both logging modules imported `logging` with both `import logging` and `from logging import ...`. Consolidated to a single `from logging import ...` block.

**Assert statement has a side-effect (1 finding)** -- `assert dict.pop(key) > 0` mutates the dict inside an assert, which is skipped when Python runs with `-O`. Fixed by extracting `pop()` to a variable before asserting.

**Unnecessary else clause in loop (1 finding)** -- `else` block after a `while` loop with no `break` is redundant. Removed the `else` and dedented the block.

**Overly complex __del__ method / resource leak bug** -- The original `__del__` used `except FileNotFoundError: return` inside each cleanup block. If the CA cert file was missing, the function returned immediately, leaving the client-cert and client-key temp files on disk. Fixed by extracting into `_cleanup_temp_files()` using independent `contextlib.suppress(OSError)` per file, so all three files are always cleaned up regardless of individual failures.

**Explicit returns mixed with implicit returns / latent type error (1 finding)** -- `all_ready()` in `probes.py` returned `None` (not `False`) when passed an unrecognised response type, violating its `-> bool` return type. Fixed by adding an explicit `return False`.

Also fixes:
- A pre-existing orphaned duplicate docstring in `src/utils/models/factory.py`
- A typo ("exeeding" -> "exceeding") in `src/initial_questions/inital_questions.py`

## Testing

No logic changes beyond bug fixes noted above. `pre-commit-check` passes (lint, typecheck, format, unit tests).

### DIFF
--- a/tests/unit/services/test_usage.py
+++ b/tests/unit/services/test_usage.py
@@ -37,7 +37,8 @@ class TestUsageTrackerCallback:
 
         # Then
         assert run_id in usage_tracker_callback.llm_start_times
-        assert usage_tracker_callback.llm_start_times.pop(run_id) > 0
+        start_time = usage_tracker_callback.llm_start_times.pop(run_id)
+        assert start_time > 0
 
     @pytest.mark.asyncio
     @pytest.mark.parametrize(
--- a/doc_indexer/src/utils/logging.py
+++ b/doc_indexer/src/utils/logging.py
@@ -1,7 +1,6 @@
-import logging
 import sys
 from logging import Logger as LoggerType
-from logging import getLogger
+from logging import StreamHandler, getLogger
 
 from pythonjsonlogger.json import JsonFormatter
 
@@ -16,13 +15,12 @@ def _configure_logging() -> None:
     _logging_configured[0] = True
 
     formatter = JsonFormatter("%(asctime)s %(name)s %(levelname)s %(message)s")
-    handler = logging.StreamHandler(sys.stdout)
+    handler = StreamHandler(sys.stdout)
     handler.setFormatter(formatter)
 
-    root = logging.getLogger()
+    root = getLogger()
     root.setLevel(LOG_LEVEL)
     root.addHandler(handler)
-
 
 
 _configure_logging()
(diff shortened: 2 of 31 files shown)


jacekon (approved):
.

tests/unit/services/test_usage.py
jacekon, line 1: feels unnecessary to do that in a test but anyway, no harm.

PR #1304, opened 2026-08-05
feat: implement response summarization in KymaReActAgent and add unit…

tanweersalah:
# Implement Response Summarization in `KymaReActAgent`

### New Feature

✨ Adds automatic response summarization to the `KymaReActAgent`'s bound tools (`kyma_query_tool`, `k8s_overview_tool`, `fetch_pod_logs_tool`). When a tool response exceeds the configured token limit, a `ToolResponseSummarizer` condenses the output before passing it to the LLM, preventing context-window overflows while preserving relevant content. If summarization fails, the agent falls back gracefully to returning the original text.

### Changes

* `src/agents/kyma/react_agent.py`:
  - Introduced `_tool_summarizer` async helper that checks token count against `TOOL_RESPONSE_TOKEN_COUNT_LIMIT`, computes the number of chunks needed, raises `TotalChunksLimitExceededError` when the chunk count exceeds `TOTAL_CHUNKS_LIMIT`, delegates to `ToolResponseSummarizer`, and falls back to plain text on generic errors.
  - Updated `_make_bound_tools` to accept a `ToolResponseSummarizer` instance and a mutable `invoke_ctx` dict, enabling tools to access the current query and `RunnableConfig` at call time.
  - Modified `kyma_query_tool`, `k8s_overview_tool`, and `fetch_pod_logs_tool` to pipe their raw output through `_tool_summarizer`.
  - Updated `KymaReActAgent.__init__` to instantiate `ToolResponseSummarizer` with the main model and maintain `_invoke_ctx`; updated `ainvoke` to populate `_invoke_ctx` with the current query and run config before each graph call.

* `tests/unit/agents/kyma/test_kyma_agent.py`:
  - Added `TestToolSummarizer` unit test class covering: short responses returned unchanged, long responses triggering the summarizer, fallback on summarizer failure, and `TotalChunksLimitExceededError` raised when chunk limit is exceeded.

* `tests/integration/agents/kyma/test_tool_summarizer.py` *(new)*:
  - Integration tests for `_tool_summarizer` validating below-limit pass-through, actual LLM-backed summarization with keyword checks, summarization quality evaluation via `GEval`, fallback behavior on various exception types, `None` config handling, query-relevance of summaries, and chunk-limit enforcement.

- [ ] 🔄 Regenerate and Update Summary

<details>
<summary>PR Bot Information</summary>

**Version:** `1.29.18`

- Event Trigger: `issue_comment.edited`
- Correlation ID: `983ea9a0-910a-11f1-85dc-78624fb8b3e7`
</details>


xincapital (dismissed):
两处问题需要在合并前修复，详见行内注释。

tests/integration/agents/kyma/test_tool_summarizer.py
```diff
@@ -0,0 +1,618 @@
...
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def summarizer(app_models) -> ToolResponseSummarizer:
+    """Create a real ToolResponseSummarizer backed by the mini model."""
+    mini_model = app_models[MAIN_MODEL_MINI_NAME]
+    return ToolResponseSummarizer(model=mini_model)
```
ruanxin: The `summarizer` fixture is backed by `MAIN_MODEL_MINI_NAME`, but `_tool_summarizer` internally calls `compute_string_token_count(text, MAIN_MODEL_NAME)` to decide whether to summarize and how many chunks to create. If the two models use different tokenizers, the chunk count used in tests will be computed against a different tokenizer than the one the production code uses, making threshold-sensitive tests like `test_boundary_exactly_at_chunk_limit_succeeds` potentially unreliable. Consider using `MAIN_MODEL_NAME` here, or add a comment explaining the intentional mismatch.

src/agents/common/conversation_summarizer.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class ConversationSummarizer:
+    """Summarize older conversation turns into a concise context paragraph."""
+
+    def __init__(self, model: IModel | Embeddings):
+        self.model = model
+
+    async def summarize(self, messages: list[BaseMessage], config: RunnableConfig | None = None) -> str:
+        """Summarize the given conversation messages into a concise recap."""
```
ruanxin, line 49: The `summarize` method does not handle an empty `messages` list. When called with `[]`, `format_conversation` returns an empty string and the LLM is still invoked with an empty conversation body, wasting a round-trip and potentially returning a confusing response. Consider short-circuiting early:

```python
if not messages:
    return ""
```

src/agents/kyma/react_agent.py
```diff
@@ -204,12 +290,54 @@ async def ainvoke(
...
         if ui_context is not None:
             human_content = f"{ui_context.as_context_message()}\n\n{query}"
-        messages = [*(chat_history or []), HumanMessage(content=human_content)]
-        payload: Any = {"messages": messages}
         run_config = RunnableConfig(callbacks=callbacks) if callbacks else None
+        prepared_history = await self._prepare_chat_history(chat_history or [], run_config)
+        messages = [*prepared_history, HumanMessage(content=human_content)]
+        payload: Any = {"messages": messages}
+        self._invoke_ctx["query"] = query
+        self._invoke_ctx["config"] = run_config
```
ruanxin: **Thread-safety hazard: `_invoke_ctx` is a shared mutable dict on the instance.**

In a FastAPI application, the same `KymaReActAgent` instance can have `ainvoke` called concurrently. A second call will overwrite `query` and `config` here while the first call's tool closures are still reading them via `invoke_ctx.get("query")`, causing tools to execute with the wrong query context.

Consider passing a per-call snapshot into `_make_bound_tools` instead of sharing `self._invoke_ctx`:

```python
# in ainvoke:
ctx_snapshot: dict[str, Any] = {"query": query, "config": run_config}
tools = [*_make_bound_tools(k8s_client, summarizer, ctx_snapshot), ...]
```

This avoids the shared mutable state entirely. Alternatively, rebuild the bound tools on each `ainvoke` call so each invocation gets its own closure scope.

ruanxin (approved):

